PR #1, opened 2017-11-17
convert library to .net standard library

brajivbala:
Converted the adyen library to a .net standard library to enable to target more .net platforms and versions.
For a a complete overview of the multi targeting capabilities , please refer to 
* https://docs.microsoft.com/en-us/dotnet/standard/net-standard

TLDR Now with these changes, anyone using the library can use it from .net 4.5 and upwards for the .net framework and .net core 2.0


AlexandrosMor:
I would like to thank you for the pull request. We tested it and the library can target multiple frameworks under the .net standard 2.0. Really great job!

brajivbala:
no problem @AlexandrosMor !
Do you have an idea when this change will be released  ?

PR #6, opened 2018-02-15
Update version number to 1.0.3

AlexandrosMor:


.gitignore
```diff
@@ -1 +1,2 @@
 .vs/
+README.md
```
lancergr: why ignore?

Issue #13, opened 2018-07-09
Disable Recurring Card


attefTouina:
Recurring service call the wrong object _listRecurringDetails in this [line ](https://github.com/Adyen/adyen-dotnet-api-library/blob/develop/Adyen.EcommLibrary/Service/Recurring.cs#L42) [the wrong endpoint] instead of _disable object

Please fix the bug

AlexandrosMor:
Hello @attefTouina 

Thank for reporting it. It is fixed now in develop branch,

Kind regards
Alexandros Moraitis

attefTouina:
Thank you for your efforts.

When this update will be published in Nuget ?

Best regards,

AlexandrosMor:
Hello @attefTouina,
 
The fix is included in version 1.0.5 and it is published on nuget.

Kind regards,
Alexandros

PR #14, opened 2018-08-10
Disable Recurring Card, Set TLS 1.2 in http 

AlexandrosMor:
Disable Recurring Card #13 
Set TLS 1.2 in http


Adyen.EcommLibrary.Test/Adyen.EcommLibrary.Test.csproj
```diff
@@ -5,11 +5,11 @@
 
     <IsPackable>false</IsPackable>
 
-    <AssemblyVersion>1.0.3</AssemblyVersion>
+    <AssemblyVersion>1.0.5</AssemblyVersion>
```
lancergr: why not 1.0.4?

PR #16, opened 2018-09-03
Docs: Replace the CSE json with payment obj

AlexandrosMor:
Replace the CSE json with payment obj. In the payment object the encrypted data passed in the additional data key value pair


rikterbeek:
HI @AlexandrosMor,

Can we combine this with terminal api release ?

Regards,
Rik
Adyen

AlexandrosMor:
Hi @rikterbeek 

it is a different topic. This has to do with the docs and it is not related to terminal api

Issue #17, opened 2018-09-14
Adding async/await to IClient interface


mwitlok:
Hi,
Have you thought to add asynchronous versions of the Request() and Post() methods of the various operations?

Currently it's possible to provide a seperate implementation of the IClient interface, but since it currently doesn't have methods that return Tasks it's difficult to use this library with an async flow from end-to-end.

AlexandrosMor:
Hi @mwitlok ,

It looks that it make sense to add asynchronous versions for Request and Post. We will add them to  the following release.

AlexandrosMor:
Hello @mwitlok,

It is added in 1.1.2 version. You can downloaded from nuget. If you have more questions please let us know

kind regards,
@AlexandrosMor

PR #24, opened 2018-10-15
Terminal api improvments

AlexandrosMor:
#17 Adding async/await to IClient interface 
Make numeric values nullable for nexo models. Numeric values are not serialized if they are not populated.


Adyen.EcommLibrary/Service/ServiceResource.cs
```diff
@@ -24,5 +25,13 @@ public string Request(string json)
             return clientInterface.Request(Endpoint, json, config);
         }
 
+
+        public async Task<string> RequestAsync(string json)
+        {
+            var clientInterface = _abstractService.Client.HttpClient;
+            var config = _abstractService.Client.Config;
+          
+            return await clientInterface.RequestAsync(Endpoint, json, config,false);
```
lancergr: let's remove await

Adyen.EcommLibrary/Service/Payment.cs
```diff
@@ -33,13 +34,30 @@ public PaymentResult Authorise(PaymentRequest paymentRequest)
             return paymentResult;
         }
 
+        public async Task<PaymentResult> AuthoriseAsync(PaymentRequest paymentRequest)
+        {
+            PaymentResult paymentResult = null;
```
lancergr:
```suggestion
            
                var jsonRequest = Util.JsonOperation.SerializeRequest(paymentRequest);
                var jsonResponse = await _authorise.RequestAsync(jsonRequest);
                return JsonConvert.DeserializeObject<PaymentResult>(jsonResponse);
```

AlexandrosMor: If we remove this try catch we need to remove it for Authorise too. I would suggest 
try { ... }catch(Exception ex) { throw ex; } similar to Authorise and Authorise3D

AlexandrosMor: If we remove this try catch we need to remove it for Authorise too. I would suggest 
try { ... }catch(Exception ex) { throw ex; } similar to Authorise and Authorise3D

AlexandrosMor: If we remove this try catch we need to remove it for Authorise too. I would suggest 
try { ... }catch(Exception ex) { throw ex; } similar to Authorise and Authorise3D

AlexandrosMor: If we remove this try catch we need to remove it for Authorise too. I would suggest 
try { ... }catch(Exception ex) { throw ex; } similar to Authorise and Authorise3D

AlexandrosMor: If we remove this try catch we need to remove it for Authorise too. I would suggest 
try { ... }catch(Exception ex) { throw ex; } similar to Authorise and Authorise3D

AlexandrosMor: If we remove this try catch we need to remove it for Authorise too. I would suggest 
try { ... }catch(Exception ex) { throw ex; } similar to Authorise and Authorise3D

Adyen.EcommLibrary/Service/ServiceResource.cs
```diff
@@ -26,12 +26,12 @@ public string Request(string json)
...
 
 
-        public async Task<string> RequestAsync(string json)
+        public Task<string> RequestAsync(string json)
         {
             var clientInterface = _abstractService.Client.HttpClient;
             var config = _abstractService.Client.Config;
           
-            return await clientInterface.RequestAsync(Endpoint, json, config,false);
+            return clientInterface.RequestAsync(Endpoint, json, config, false);
```
AlexandrosMor: Hey @lancergr, The Await Async is removed for RequestAsync

PR #29, opened 2018-10-18
Issue/local api endpoint

AlexandrosMor:
Added local api endpoint.
Fix encryption serialization 
Return null when response is "ok" for terminal api


Adyen.EcommLibrary.Test/BaseTest.cs
```diff
@@ -150,6 +150,7 @@ protected Client CreateMockTestClientApiKeyBasedRequest(string fileName)
         protected Client CreateMockTestClientPosApiRequest(string fileName)
         {
             var config = new Config();
+            config.Endpoint = @"http://dummylocalterminalapi:8443";
```
lancergr:
```suggestion
            config.Endpoint = @"https://_terminal_:8443/nexo/";
```

PR #30, opened 2018-10-22
Issue/local api endpoint

AlexandrosMor:
Correct receipt field value to text, 
Exception RequestAsync, 
Remove await from RequestAsync ServiceResource
Rename the dummy terminal api url 
Http client skip certificate validation in case local api


Adyen.EcommLibrary/Config.cs
```diff
@@ -18,8 +18,10 @@ public class Config
         public string HppEndpoint { get; set; }
         public string SkinCode{ get; set; }
         public string HmacKey { get; set; }
+
         //Terminal cloud api
         public string XApiKey { get; set; }
         public string CloudApiEndPoint { get; set; }
+        public bool SkipCertValidation { get; set; } = true;
```
lancergr: let's default to false here

PR #33, opened 2018-11-07
Async changes and removed expect header

markybry:
* Moved the await for reader.ReadToEndAsync so that it is inside the response using statement.  Just in case it is disposed before the IO operation completes.
* responseTextTask.Result could have caused a thread lock.
* removed if (response.StatusCode != HttpStatusCode.OK) as this was happening after response.GetResponseStream() which will throw an exception for non 200 Status codes.

* Removed expect header as this is handled by System.Net and should be configured if required through ServicePoint.Expect100Continue . This is true by default which is the desired behavior I believe.
https://docs.microsoft.com/en-us/dotnet/api/system.net.servicepoint.expect100continue?view=netframework-4.7.2


Adyen.EcommLibrary/HttpClient/HttpURLConnectionClient.cs
```diff
@@ -79,21 +71,23 @@ public async Task<string> RequestAsync(string endpoint, string json, Config conf
...
+                    using (var reader = new StreamReader(response.GetResponseStream(), _encoding))
+                    {
+                        responseText = await reader.ReadToEndAsync();
+                    }
                 }
             }
-            return await responseTextTask;
+            catch (WebException e)
+            {
+                Console.WriteLine(e);
```
lancergr: To avoid printing on stdout, I believe is better to log this by using the client logger if it is set: https://github.com/Adyen/adyen-dotnet-api-library/blob/63eda4a1a379a2605d3f9247e128edd632013041/Adyen.EcommLibrary/Client.cs#L96

markybry: True. It ended being included when I copied across from the non async version of Request to make them the same. I can update use Logline in this PR for both.

markybry: Removed the try catch and logging as this should be done higher up in the abstract service or just leave it to the consumer of the library in my opinion. What do you think.

AlexandrosMor: Yes it is true. The try catch can be removed as the exceptions are handled in higher level

PR #37, opened 2018-11-30
Feature/pw 710 checkout

AlexandrosMor:
v40 checkout api methods
/paymentMethods
/payments
/payments/details


Adyen.EcommLibrary/Constants/ClientConfig.cs
```diff
@@ -6,12 +6,15 @@ public class ClientConfig
         public static string EndpointLive = "https://pal-live.adyen.com";
         public static string HppTest = "https://test.adyen.com/hpp";
         public static string HppLive = "https://live.adyen.com/hpp";
+        public static string CheckoutEndpointTest = "https://checkout-test.adyen.com";
+        public static string CheckoutEndpointLive = "https://checkout-live.adyen.com";
```
Aleffio: Please use the prefix-suffix structure for LIVE we are using in the other libraries:
e.g.: https://github.com/Adyen/adyen-php-api-library/blob/d22384bb9b2d1d7dfa57e504697f3c58bb4abf19/src/Adyen/Client.php#L30

PR #44, opened 2018-12-23
Pw 876

AlexandrosMor:
Deserialize RepeatedresponseMessageBody messagebody


AlexandrosMor:
@rkewlani 
1. RepeatedMessageResponse is optional but if it is not null then RepeatedResponseMessageBody is required.
2.  Fixed
3. Add dynamic casting

PR #56, opened 2019-02-06
Change ShopperInteraction to allow Moto PaymentRequests

Pearson0123:
Changed the ShopperInteraction property to be nullable enabling moto
payment requests to be serialized by StringEnumConverter correctly.

This PR is only the fix change. I have a unit test on another branch here (https://github.com/Pearson0123/adyen-dotnet-api-library/blob/shopperInteractionTest/Adyen.EcommLibrary.Test/PaymentTest.cs) if that is required.

To reproduce this issue you can run the unit test (TestShopperInteractionSerialization) on develop.


subnormalseries:
This failed the build because the Test I added in my pull request checked that the serialized value for Ecommerce would be empty. Whereas setting the enum to nullable meant the default value for the enum was null and thus every shopper interaction would be serialized and thus not empty.

Options are:
Revert either commit
OR 
Remove failing unit test as with both PR merged it no longer makes sense.

AlexandrosMor:
Yes indeed,
I run the tests and it is failing. We will revert the change. 
Thank you for noticing it 
kind regards,
Alexandros
Adyen

PR #58, opened 2019-02-11
Add Idemopotency-Key to header of web request, allowing duplicate requests to be recognised

haydnchapman:
Support for API Idempotency included in this PR.

Currently there is no way to specify the idempotent key in the header using the client. This allows an optional key to be passed in when calling the Authorise methods.

Functionality from Adyen docs: [API idempotency](https://docs.adyen.com/developers/development-resources/api-idempotency)


haydnchapman:
Resolved merge conflicts

AlexandrosMor:
Great work. Thank you for your effort. We will merge the code soon.

Adyen.EcommLibrary/HttpClient/HttpURLConnectionClient.cs
```diff
@@ -60,14 +60,14 @@ public string Request(string endpoint, string json, Config config, bool isApiKey
         /// <param name="config"></param>
         /// <param name="isApiKeyRequired"></param>
         /// <returns>Task<></returns>
-        public async Task<string> RequestAsync(string endpoint, string json, Config config, bool isApiKeyRequired)
+        public async Task<string> RequestAsync(string endpoint, string json, Config config, bool isApiKeyRequired, string idempotencyKey = null)
```
rkewlani: This approach is not backward compatible because RequestAsync with 3 parameters will not be supported and some merchants may not want to use idempotency key.

haydnchapman:
The idempotencyKey is an optional parameter, so existing calls to
RequestAsync will not require updating (as null is specified as the default
value), so if a merchant doesn't want to use the idempotency key, they
don't need to specify it. There are unit tests to ensure if the key isn't
supplied, the header isn't added.

AlexandrosMor:
Hello @haydnchapman,
Thank you for your PR. We are trying to keep all the adyen's SDKs aligned. Your approach for the idempotency key is different from what we implement for [adyen-java-api-library](https://github.com/Adyen/adyen-java-api-library). The implementation for java can be found [here](https://github.com/Adyen/adyen-java-api-library/pull/129/files). We decided to use a RequestOptions class rather than single property so the parameters are more flexible.
Is it possible for you to refactor the code to a similar implementation ?  

Kind regards, 
Alexandros 
Adyen

haydnchapman:
@AlexandrosMor - yep that's no problem. I will find the time over the next day or two to make the changes and update this PR.

Thanks,
Haydn

haydnchapman:
Hi @AlexandrosMor 

I have made the necessary changes to bring the change more inline with the Java SDK.

Thanks,
Haydn

AlexandrosMor:
Hey @haydnchapman,

Thank you for your PR we will review it and merge it as soon as possible.

Kind regards,
Alexandros
Adyen

Adyen.EcommLibrary/HttpClient/HttpURLConnectionClient.cs
```diff
@@ -57,15 +58,16 @@ public string Request(string endpoint, string json, Config config, bool isApiKey
         /// <param name="json"></param>
         /// <param name="config"></param>
         /// <param name="isApiKeyRequired"></param>
+        /// <param name="idempotencyKey">Optional parameter used to specify the idemopotency key for the request</param>
```
AlexandrosMor: Minor : This should be Request options

AlexandrosMor: Minor: This should be requestOptions

haydnchapman: @AlexandrosMor - good spot, updated

PR #60, opened 2019-02-13
update checkout to v41

AlexandrosMor:


rikterbeek:
@AlexandrosMor update the version

PR #102, opened 2019-06-06
Rename the library Adyen.EcommLibrary to Adyen

AlexandrosMor:
Renaming the library projects  
Adyen.EcommLibrary -> Adyen
Adyen.EcommLibrary.Test -> Adyen.Test
Adyen.EcommLibrary.sln -> Adyen.sln


rikterbeek:
Hi @AlexandrosMor just want to make sure you have moved the files rather then removed and added. Do we keep the git history on files

PR #106, opened 2019-06-11
Refactoring

AlexandrosMor:
1 Correct recurring, 
2 httpclientTimeout to httpRequestTimeout, 
3 Empty string check for serialization
4 Regenerate the nexo models
5 Add unit test for event notification


Adyen.Test/CloudApiPosRequestTest.cs
```diff
@@ -57,7 +57,7 @@ public void TestCloudApiTransactionStatusResponseSuccess()
 
             try
             {
-                var transactionStatusResponse = (TransactionStatusResponse)saleToPoiResponse.Item;
+                var transactionStatusResponse = (TransactionStatusResponse)saleToPoiResponse.MessagePayload;
```
rikterbeek: do we do the same for our java lib ?

AlexandrosMor: In the dotnet the Item is an object which is a placeholder for the any of the response. We rename it to MessagePayload as it is more clear. In java there are all the responses as members of the SaleToPOIResponse class.

Issue #109, opened 2019-06-19
Not passing 'liveEndpointUrlPrefix' to client in live environment results in ambiguous error


GonathanJawecki:
Hey, just recently encountered this small issue when I was using your library.

When newing up a Client class I was using the constructor: 
```c# 
Client(string xapikey, Environment environment) 
```
for both Test and Live environments, this worked fine for Test.

However, for Live I was getting this error when I was calling the `Checkout.PaymentMethods` method:

```c#
System.UriFormatException: Invalid URI: The format of the URI could not be determined.
   at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)
   at System.Uri..ctor(String uriString)
   at System.Net.WebRequest.Create(String requestUriString)
   at Adyen.EcommLibrary.HttpClient.HttpUrlConnectionClient.GetHttpWebRequest(String endpoint, Config config, Boolean isApiKeyRequired, RequestOptions requestOptions)
   at Adyen.EcommLibrary.HttpClient.HttpUrlConnectionClient.Request(String endpoint, String json, Config config, Boolean isApiKeyRequired, RequestOptions requestOptions)
   at Adyen.EcommLibrary.Service.ServiceResource.Request(String json, RequestOptions requestOptions)
   at Adyen.EcommLibrary.Service.Checkout.PaymentMethods(PaymentMethodsRequest paymentMethodsRequest)
```
Obviously this was because I wasn't using the constructor:
```c# 
Client(string xapikey, Environment environment, string liveEndpointUrlPrefix) 
``` 
to declare the data center correctly for Live. However, this wasn't picked up until we released to our Staging environment. As the Live URL was malformed since it was missing the `liveEndpointUrlPrefix`.

I feel that ideally it would be nice to have had a more descriptive error before the code made it to a method call. I don't really understand the reason behind the config value not setting the `CheckoutEndpoint` if the `liveEndpointUrlPrefix` is null in this switch statement? is this Possibly a bug?

```c#
switch (environment)
{
    case Environment.Test:
        Config.Endpoint = ClientConfig.EndpointTest;
        Config.HppEndpoint = ClientConfig.HppTest;
        Config.CloudApiEndPoint = ClientConfig.CloudApiEndPointTest;
        Config.CheckoutEndpoint = ClientConfig.CheckoutEndpointTest;
        break;
    case Environment.Live:
        Config.Endpoint = ClientConfig.EndpointLive;
[...]

PR #112, opened 2019-06-20
Added exception for invalid usage of live client constructor generation

GonathanJawecki:
Not 100% sure if you're happy with this change, basically made it throw an `InvalidOperationException` if `liveEndpointUrlPrefix` is empty when setting/generating the live URLs in the Client constructor. I'm going under the assumption that it should always be populated for Live environments.

-Added 2 unit tests.
-Edited 1 unit test.
-Throw an `InvalidOperationException` if `liveEndpointUrlPrefix` is empty on Live environment.

### DIFF
--- a/Adyen.Test/CheckoutTest.cs
+++ b/Adyen.Test/CheckoutTest.cs
@@ -11,14 +11,42 @@ namespace Adyen.Test
     [TestClass]
     public class CheckoutTest : BaseTest
     {
+        /// <summary>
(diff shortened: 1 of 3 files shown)


GonathanJawecki:
Relates to this Issue https://github.com/Adyen/adyen-dotnet-api-library/issues/109

AlexandrosMor:
Fixes #109

PR #127, opened 2019-07-23
 [PW-1399] Add serialize test for AuthenticationMethodType.OnlinePIN

AlexandrosMor:


AlexandrosMor:
Serialization handles both elements "OnLinePIN" and "OnlinePIN"

PR #128, opened 2019-07-24
[PW-1456] Refactor names and add the ForceEntryMode enum

AlexandrosMor:


AlexandrosMor:
Add ForceTypeMode enum, rename Instalment1 to Instalment and fix name from Type to TypeCode for CheckTypeCodeType

Issue #115, opened 2019-06-21
Library compatibility with API v49


BenjaminFeenstra:
Implementing the new [drop-in solution](https://docs.adyen.com/checkout/drop-in-web) requires API version 49. This .NET library still targets v40 however and renders it incompatible with the drop-in solution. This is because the `/payments` endpoint response has been changed in v49.

FrancisYergeau:
I'm working to install v3.0.2 to use the Drop-In and the API Version is still "v40" I get a Timeout is it the reason?

FrancisYergeau:
I've contacted the support they tell me that the URL is wrong when trying to get the list of paymentMethods:
----
https://checkout-test.adyen.com/checkout//checkout/PaymentSetupAndVerification/paymentMethods/v49 (wrong)
https://checkout-test.adyen.com/checkout/paymentMethods/v49 (correct)
----

Can someone check why?

rikterbeek:
Hi @BenjaminFeenstra,

Thanks for your comment. We will investigate what changes this will bring across  the models. We let you know. For a short term fix you can manually update the version.

Regards,
Rik
Adyen

mattsage1991:
Is there any updates on this? I'm having the same problem where I'm trying to send a payment through the /payments endpoint but the API version in the client library is still v40.

yang-xiaodong:
We also encountered the same problem, I can't find the `action` object in the returned result.

![image](https://user-images.githubusercontent.com/8205994/61436550-9e11b200-a96d-11e9-86e6-7edd336a161e.png)
![image](https://user-images.githubusercontent.com/8205994/61436561-a1a53900-a96d-11e9-9808-fcaa2ef8df88.png)

mattsage1991:
after doing a few tests I realised that the action only gets returned when the api version is set to v49. Anything lower and no action is returned.

mattsage1991:
@yang-xiaodong I ended up cloning the repository and changing the ApiVersion in the ConfigConstants.cs in the end. If you don't want to do that, you could always send the request yourself via HttpClient as you mentioned.

Recio:
Just updated to 3.2.0
Action property is still not available on the response object.

Fiddler shows the comm going to v49 and returning an action object. If you're missing this feature don't update just yet

AlexandrosMor:
Hello @Recio 
Can you check this PR? https://github.com/Adyen/adyen-dotnet-api-library/pull/132

[...]

Issue #108, opened 2019-06-14
Missing Result Codes for Checkout API


jaw64:
The [documentation for the Checkout API](https://docs.adyen.com/checkout/payment-result-codes/) lists the following as possible result codes for the `/payments` and `/payments/details` endpoints:

- `AuthenticationFinished`
- `Authorised`
- `Cancelled`
- `ChallengeShopper`
- `Error`
- `IdentifyShopper`
- `Pending`
- `PresentToShopper` (not included in the referenced doc, but I have seen it mentioned elsewhere on the site)
- `Received`
- `RedirectShopper`
- `Refused`

The result codes currently in the [Checkout model `PaymentsResponse.ResultCodeEnum`](https://github.com/Adyen/adyen-dotnet-api-library/blob/cb07fd51cd9e5e6827f461eda1a485b07a10da6a/Adyen/Model/Checkout/PaymentsResponse.cs#L24) are as follows:

- `Authorised`
- `Cancelled`
- `ChallengeShopper`
- `Error`
- `IdentifyShopper`
- `PartiallyAuthorised`
- `PresentToShopper`
- `Received`
- `RedirectShopper`
- `Refused`

Most of documented result codes are included in this .NET SDK (and even an extra one which is not documented -- i.e. `PartiallyAuthorised`) but there are some discrepancies between the two. All of the documented result codes for Checkout API should be in the `PaymentsResponse.ResultCodeEnum`.

rikterbeek:
Hi @jaw64,

The list here could reflect checkout and classic integration.

We have as well this enumClass for classic integration: https://github.com/Adyen/adyen-dotnet-api-library/blob/cb07fd51cd9e5e6827f461eda1a485b07a10da6a/Adyen/Model/Enum/ResultCodeEnum.cs


Regards,
Rik
Adyen

jaw64:
Hi @rikterbeek,

The enum you've referenced is also missing the `Pending` result code.

Regards,
Jordan

AlexandrosMor:
Hello @jaw64,

We are going to review the ResultCodeEnum soon as we are going to move the library to api version v49

PR #130, opened 2019-07-30
[PW-1414]Update to version v49

AlexandrosMor:


AlexandrosMor:
fixes #108 
fixes #115

PR #136, opened 2019-08-05
[PW-1583] Add Sepa fields and Bank Account

AlexandrosMor:
Add sepa fields sepa.owner, sepa.ibanNumber and BankAccount


Adyen/Model/Checkout/DefaultPaymentMethodDetails.cs
```diff
@@ -90,7 +96,115 @@ public string ToJson()
         /// <returns>Boolean</returns>
         public override bool Equals(object input)
         {
-            return this.Equals(input as DefaultPaymentMethodDetails);
+            return this.Equals(input as Configuration);
```
Aleffio:
```suggestion
            return this.Equals(input as DefaultPaymentMethodDetails);
```

AlexandrosMor: @Aleffio Fixed 👍

PR #150, opened 2019-08-22
a proposed change to GetDataToSign function in the HMACvalidator

robertftenbosch:
In the example on the website the boolean is shown with a lowercase t. 
I presume (maybe mistakenly) that the encyption done on the adyenside is done with a lower case t(rue) or f(alse)
Doing this in dotnet will result in a string with an uppercase T or F.  (that why is propose a ToLower()
rendering a different hash and will give a false negative.
Shouldn't this be the same as in the java test?
because your are using the same data and same key?
in the java test:
`assertEquals("pspReference:originalReference:merchantAccount:reference:1000:EUR:EVENT:true", data)`
[java](https://github.com/Adyen/adyen-java-api-library/blob/develop/src/test/java/com/adyen/UtilTest.java#L111)
int the nodejstest:
`expect(data).toEqual("pspReference:originalReference:merchantAccount:reference:1000:EUR:EVENT:true");`
[nodejs](https://github.com/Adyen/adyen-node-api-library/blob/develop/src/__tests__/hmacValidator.spec.ts#L42
)

### DIFF
--- a/Adyen/Util/HMACValidator.cs
+++ b/Adyen/Util/HMACValidator.cs
@@ -95,7 +95,7 @@ namespace Adyen.Util
                 Convert.ToString(amount.Value),
                 amount.Currency,
                 notificationRequestItem.EventCode,
-                notificationRequestItem.Success.ToString()
+                notificationRequestItem.Success.ToString().toLower()
             };
             return String.Join(":", signedDataList);
         }
--- a/Adyen/Util/HMACValidator.cs
+++ b/Adyen/Util/HMACValidator.cs
@@ -95,7 +95,7 @@ namespace Adyen.Util
                 Convert.ToString(amount.Value),
                 amount.Currency,
                 notificationRequestItem.EventCode,
-                notificationRequestItem.Success.ToString().toLower()
+                notificationRequestItem.Success.ToString().ToLower()
             };
             return String.Join(":", signedDataList);
         }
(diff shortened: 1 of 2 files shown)


Adyen/Util/HMACValidator.cs
```diff
@@ -95,7 +95,7 @@ public string GetDataToSign(NotificationRequestItem notificationRequestItem)
                 Convert.ToString(amount.Value),
                 amount.Currency,
                 notificationRequestItem.EventCode,
-                notificationRequestItem.Success.ToString()
+                notificationRequestItem.Success.ToString().ToLower()
```
rikterbeek: why do we need to lowercase this ?

robertftenbosch: Again, I could be mistaken but I believe that the encryption done on the adyenside in done with a lowercase t or f. 
At least thats how the example on the website [docs](https://docs.adyen.com/development-resources/notifications/verify-hmac-signatures) is and how the tests in the other sdks expect the string to be.  
and this is where I got a false negative when I was implementing this code. changing the boolean toString with toLower fixed it. Because this wil result in a "true" string instead of a "True" string.

rikterbeek: @robertftenbosch thanks your right!

Recio: Also validated on my side, for the signatures to match I had to ToLower() on the Success parameter, I ended up copy pasting the implementation to my project (to pull the fields and concat them, the function is not available on 3.2.3) and came across the same issue.

PR #177, opened 2019-10-11
[PW-1659] Add brands field

AlexandrosMor:


Adyen.Test/CheckoutTest.cs
```diff
@@ -216,6 +216,23 @@ public void PaymentMethodsTest()
             Assert.AreEqual(paymentMethodsResponse.PaymentMethods.Count, 65);
 
         }
+
+        [TestMethod]
```
rkewlani: Please add negative test case as discussed.

Adyen/Model/Checkout/PaymentMethod.cs
```diff
@@ -47,7 +47,7 @@ public partial class PaymentMethod :  IEquatable<PaymentMethod>, IValidatableObj
         /// <param name="PaymentMethodData">Echo data required to send in next calls..</param>
         /// <param name="SupportsRecurring">Indicates whether this payment method supports tokenization or not..</param>
         /// <param name="Type">The unique payment method code..</param>
```
Aleffio:
```suggestion
        /// <param name="Type">The unique payment method code..</param>
        /// <param name="Brands">List of possible brands. For example: visa, mc.</param>
```
or something like that

PR #187, opened 2019-11-08
Addition of all documented notification event codes

bomehlsen:
All existing notification event codes should exist in the library.

The `all documented notification event codes` mentioned in the commit message are from here: https://docs.adyen.com/development-resources/notifications/understand-notifications


AlexandrosMor:
Hello @bomehlsen, 

Thank you for the PR. We are going to review it and merge it as soon as possible.

Kind regards,
Alexandros Moraitis

PR #195, opened 2019-12-18
Fix for Android Drop-In 3.5.1+ compatibility

ChrisHG-Ticketer:
Deprecate RecurringDetailReference property (but retain for backward compatibility) and include StoredPaymentMethodId for forward compatibility (Android Drop-In version 3.5.1+ requires this field)


AlexandrosMor:
Hello @ChrisHG-Ticketer 

Thank you for your pull request. This change is part of the checkout api version v51. We are going to update the library to v51 soon so we will include your pull request too. 

kind regards,
Alexandros

PR #234, opened 2020-02-21
Create dotnetcore.yml

AlexandrosMor:
test the action for building the library


rikterbeek:
@AlexandrosMor looks like more changes can you explain why other code is changed ? Let's fix the conflicts.

PR #236, opened 2020-03-05
[PW-2126]Change MenuEntryMenuEntry from integer to array

### LINKED ISSUE #230: Collect user input issue (MenuButtons): a display response containing an array as a value cannot be deserialized due to a bug in Newtonsoft.Json bug
A display response containing an array as a value cannot be deserialized due to a bug in Newtonsoft.Json described here [https://github.com/JamesNK/Newtonsoft.Json/issues/2273](https://github.com/JamesNK/Newtonsoft.Json/issues/2273)

AlexandrosMor:
fixes #230


Adyen.Test/TerminalApiInputTest.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace Adyen.Test
+{
+    [TestClass]
+    public class TerminalApiInputTest : BaseTest
+    {
+        [TestMethod]
+        public void InpturequestTest()
```
Aleffio:
```suggestion
        public void InputRequestTest()
```

PR #250, opened 2020-04-04
Changed ctor for Name model to not (apparently) accept nulls and made props readonly

JoachimL:
We recently integrated against Adyen, and experienced a couple of unnecessary issues which would have been avoided with a more intuitive code interface.

This PR alters the Name object to not (apparently) allow nulls. The original code exposes a ctor which allows not passing in arguments, but immediately fails if firstName or lastName are null. It seems a bit silly to set default values which immediately throws exceptions.

The PR also removed the check for gender == null, since enums can't actually be null, throws ArgumentNullException instead of InvalidDataExeption and renames parameters to pascal case, since that's conventional .net.

I'm not sure if the checks should actually be string.IsNullOrEmpty, though, but this should keep the original functionality while giving devs a better experience.

Lastly,  all properties are changed to readonly, since they are required to construct the object.

### DIFF
--- a/Adyen/Model/Checkout/Name.cs
+++ b/Adyen/Model/Checkout/Name.cs
@@ -24,7 +24,6 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
-using System.IO;
 using System.Runtime.Serialization;
 using System.Text;
 using Newtonsoft.Json;
@@ -122,14 +121,14 @@ namespace Adyen.Model.Checkout
         /// </summary>
         /// <value>The name&#39;s infix, if applicable. &gt;A maximum length of twenty (20) characters is imposed.</value>
         [DataMember(Name = "infix", EmitDefaultValue = false)]
-        public string Infix { get; set; }
+        public string Infix { get; }
 
         /// <summary>
         /// The last name.
         /// </summary>
         /// <value>The last name.</value>
         [DataMember(Name = "lastName", EmitDefaultValue = false)]
-        public string LastName { get; set; }
+        public string LastName { get;  }
 
         /// <summary>
         /// Returns the string presentation of the object
--- a/Adyen/Model/Checkout/Name.cs
+++ b/Adyen/Model/Checkout/Name.cs
@@ -36,7 +36,7 @@ namespace Adyen.Model.Checkout
     /// Name
     /// </summary>
     [DataContract]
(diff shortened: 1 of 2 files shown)


AlexandrosMor:
Hello @JoachimL,
 
Thank you for your PR, We will review it and let you know as soon as possible.

kind regards,
Alexandros

Adyen/Model/Checkout/Name.cs
```diff
@@ -79,63 +78,57 @@ public enum GenderEnum
...
-        /// <param name="FirstName">The first name. (required).</param>
-        /// <param name="Gender">The gender. &gt;The following values are permitted: &#x60;MALE&#x60;, &#x60;FEMALE&#x60;, &#x60;UNKNOWN&#x60;. (required).</param>
-        /// <param name="Infix">The name&#39;s infix, if applicable. &gt;A maximum length of twenty (20) characters is imposed..</param>
-        /// <param name="LastName">The last name. (required).</param>
-        public Name(string FirstName = default(string), GenderEnum Gender = default(GenderEnum), string Infix = default(string), string LastName = default(string))
+        /// <param name="firstName">The first name. (required).</param>
+        /// <param name="gender">The gender. &gt;The following values are permitted: &#x60;MALE&#x60;, &#x60;FEMALE&#x60;, &#x60;UNKNOWN&#x60;. (required).</param>
+        /// <param name="infix">The name&#39;s infix, if applicable. &gt;A maximum length of twenty (20) characters is imposed..</param>
+        /// <param name="lastName">The last name. (required).</param>
+        public Name(string firstName, GenderEnum gender, string infix, string lastName)
```
AlexandrosMor, line 85: infix is not required field so it can be string Infix = default(string) [you can check it here in docs](https://docs.adyen.com/api-explorer/#/PaymentSetupAndVerificationService/v52/payments__reqParam_shopperName)

JoachimL: The problem with this is that you can't have required parameters _after_ optional parameters :-/ Adding a default to infix wouldn't compile without defaulting lastname as well.
Ideally, the ctor should be (firstname, lastname, gender, infix = null), but to adjust the order of parameters would be a very breaking change and could potentially trip up existing implementations.

AlexandrosMor: Yes, good catch. The costructor can be as you set it

PR #255, opened 2020-04-30
[PW-2336] Add paymentMethodDetails classes 

AlexandrosMor:
- Add payment method classes based on open api specs.
- Update the payment request by adding payment method interface
- Add basic unit tests


Adyen.Test/PaymentMethodDetailsTest.cs
```diff
@@ -0,0 +1,147 @@
...
+{
+    [TestClass]
+   public class PaymentMethodDetailsTest
+    {
+        [TestMethod]
+        public void TestAchPaymentMethod()
+        {
+            var paymentRequest = new PaymentRequest
+            {
+                MerchantAccount = "YOUR_MERCHANT_ACCOUNT",
```
KadoBOT, line 36: What about making this a constant? So if in the future you want to make this an e2e test, you change the merchant account in one place only.

Adyen/Model/Checkout/AchDetails.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Runtime.Serialization;
+using System.Text;
+using Newtonsoft.Json;
+
+namespace Adyen.Model.Checkout
+{
+
+    /// <summary>
+    /// 
+    /// </summary>
```
KadoBOT, line 31: Since there's no summary, this can be removed.

PR #263, opened 2020-05-14
Use cryptographically secure PRNG for IV mod

pavisalavisa:
`System.Random` is not cryptographically secure as pointed out [here](https://codingvision.net/security/c-predict-random-number-generator-net):

> If we manage to leak a continuous set of 55 generated numbers, we have enough information to describe and construct a new generator (by providing a circular array of states) which will output the same numbers as the original but can be used as a predictor.

AES in CBC mode with predictable IVs is prone to chosen plaintext attack and as such should be avoided.

An alternative would be to use cryptographically secure pseudorandom number generator that's implemented in `RNGCryptoServiceProvider`.

### DIFF
--- a/Adyen/Security/IvModGenerator.cs
+++ b/Adyen/Security/IvModGenerator.cs
@@ -21,7 +21,7 @@
 //  */
 #endregion
 
-using System;
+using System.Security.Cryptography;
 
 namespace Adyen.Security
 {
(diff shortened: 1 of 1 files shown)


AlexandrosMor:
Hello @pavisalavisa 
Thank you for your pr. We will review it as soon as possible
regards,
Alexandros

PR #264, opened 2020-05-18
Add GitHub actions for testing

AlexandrosMor:
**Description**
This pr is only the yml file for github actions. Warnings and others will be fixed in another PR 

**Tested scenarios**
Github actions run the tests

### DIFF
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        dotnet: [ '2.0.0','2.2.103', '3.1.100' ]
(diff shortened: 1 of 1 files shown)


.github/workflows/dotnetcore.yml
```diff
@@ -0,0 +1,19 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup .NET Core
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 2.2.108
```
rikterbeek, line 15: please include multiple dot-net versions that we want to support.

AlexandrosMor: I removed the 3 dotnet core version. We can investigate adding .net core 3 and windows framework when we pick up this ticket. For now it is similar to travis.

PR #266, opened 2020-05-19
Add feature and bug report template

AlexandrosMor:


AlexandrosMor:
The templates are going to be standard across Adyen's repositories. This pr is not relevant anymore

PR #269, opened 2020-05-25
[PW-2455] Adding sdkData in action response model

AlexandrosMor:
**Description**
Adding sdkData in the CheckoutPaymentsAction class.

**Tested scenarios**
Tested with paypal response which includes the sdk data


Adyen/Model/Checkout/CheckoutPaymentsAction.cs
```diff
@@ -451,6 +460,11 @@ public bool Equals(CheckoutPaymentsAction input)
                     (this.ShopperEmail != null &&
                     this.ShopperEmail.Equals(input.ShopperEmail))
                 ) &&
+                (
+                    this.SdkData == input.SdkData ||
+                    (this.SdkData != null &&
+                    this.SdkData.Equals(input.SdkData))
+                ) &&
```
Aleffio, line 467: can you please add sdkData to the GetHashCode method?

AlexandrosMor: Done

PR #286, opened 2020-06-26
Added GetTerminalDetails endpoint

karlkaal:
## Tested scenarios
My main motivation for using it is to retrieve the local IP address of the terminal for use with local communication instead of cloud based. 

Not sure about possible fields in the response data, so it's just implemented as a dictionary.

### DIFF
--- a/Adyen/Model/PosTerminalManagement/GetTerminalDetailsResponse.cs
+++ b/Adyen/Model/PosTerminalManagement/GetTerminalDetailsResponse.cs
@@ -1,0 +1,42 @@
+﻿using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+using System.Text;
+
+namespace Adyen.Model.PosTerminalManagement
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    [DataContract]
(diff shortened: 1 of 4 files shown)


Adyen/Model/PosTerminalManagement/GetTerminalDetailsResponse.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Runtime.Serialization;
+using System.Text;
+
+namespace Adyen.Model.PosTerminalManagement
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    [DataContract]
+    public class GetTerminalDetailsResponse : Dictionary<string, object>
```
AlexandrosMor, line 13: Response should be based response parameters and not key value pair. Please check the [documentation] (https://docs.adyen.com/api-explorer/#/postfmapi/v1/post/getTerminalDetails__section_resParams)

AlexandrosMor:
Hello @karlkaal,

Would you have some time to make the changes on this pr? 

Kind regards,
Alexandros

AlexandrosMor:
I created another [PR](https://github.com/Adyen/adyen-dotnet-api-library/pull/308) adding the classes generated by the open api specs. Furthermore a unit test is added.

Issue #290, opened 2020-06-30
Feature: paymentMethod under PaymentsResponse


Recio:
Hi, 

long time no see, it's great to see the updates that have been made to the lib so far :)

While integrating 2 more payment methods on our system (vipps and trustly) we noticed that there is no additionalData being returned (both on the api and on the lib, verified by fiddler) when calling payments/details.

Reaching out to the integration team they brought to my attention that I should look for the field paymentMethod
This field is not showing up both on the lib or api documentation (https://docs.adyen.com/api-explorer/#/PaymentSetupAndVerificationService/v52/post/payments/details), this field is visible when inspecting the traffic through fiddler.

What are the best practices regarding this? can I open a PR to have it on the lib or should I extend it on my side for our integration?

Thanks!

AlexandrosMor:
Hello @Recio,

Thank you for the nice comments. 
 If I understand correct there is a missing field from payments details response which returns the additional data. 
Is this field in the json response from the api? Yes you can open a pull request and please include a unit test. 

Kind regards,
Alexandros Moraitis, 
Adyen

AlexandrosMor:
Hello @Recio,

Talking with the checkout team there is a plan in the future to change the way that we handle the payment method in the payments response. For now the recommended way is to retrieve the payment method in the additional data from the payments response. For example:

```
{
    "additionalData": {
        "fraudResultType": "GREEN",
        "authCode": “1234”,
         "cardHolderName": "John Smith",
        "cardSummary": “999”,
        "fraudManualReview": "false",
        "cardIssuingBank": "ADYEN TEST BANK",
        "expiryDate": "10/2020",
        "cvcResult": "1 Matches",
        "cardBin": “009999”,
        "aliasType": "Default",
        "alias": “H0000000000”,
        "paymentMethod": "visa",
        "cardPaymentMethod": "visa",
        "paymentMethodVariant": "visa",
        "merchantReference": "Your order number",
        "cardIssuingCountry": "NL"
    },
    "pspReference": “12312321321312”,
    "resultCode": "Authorised",
    "merchantReference": "Your order number"
}
```
For that reason we will not merge you pull request.
If you have more questions please let us know

kind regards,
Alexandros

PR #301, opened 2020-07-23
Adds paymentMethod as a property for PaymentsResponse

Recio:
## Summary
Adds paymentMethod as a property for PaymentsResponse
This field is particularly useful to determine which payment method the customer verified without having to resort to storing and manipulating temporary data from the submit request.
There is still no documentation available on the Adyen Docs.

## Tested scenarios
Basic deserialization of a successful request (performed locally and exported to .json)

Let me know if there's other tests you find necessary for this approval

**Fixed issue**:  #290


AlexandrosMor:
As we mentioned in the issue #290 the recommended way to retrieve the payment method is from the additional data from the payments response.

PR #309, opened 2020-08-14
[PW-2463] Fix the warnings in the dot net library

AlexandrosMor:
## Summary
Fixed the warnings in the dot net library. Most of the fixes are:
- Suppress the warnings for using obsolete fields which are not removed yet
- Warnings on comparisons that the result of the expression is always 'true' since a value of type is not nullable similar to int, enum, bool.
- Warnings for classes that overrides Object.Equals(object o) but does not override Object.GetHashCode() 

## Tested scenarios
All unit test pass

**Fixed issue**:


Adyen/Model/MarketPay/DocumentDetail.cs
```diff
@@ -150,6 +142,7 @@ public DocumentDetail(string accountHolderCode = default(string), string bankAcc
                 this.Filename = filename;
             }
             this.DocumentType = documentType;
+            this.DocumentType = documentType;
```
Aleffio, line 145: why? :)
```suggestion
```

AlexandrosMor: ops...

Adyen/Model/BrowserInfo.cs
```diff
@@ -199,16 +199,11 @@ public override int GetHashCode()
...
-                    hash = hash * 59 + this.ScreenWidth.GetHashCode();
-                if (this.TimeZoneOffset != null)
-                    hash = hash * 59 + this.TimeZoneOffset.GetHashCode();
-                if (this.JavaEnabled != null)
-                    hash = hash * 59 + this.JavaEnabled.GetHashCode();
+                hash = hash * 59 + this.ColorDepth.GetHashCode();
+                hash = hash * 59 + this.ScreenHeight.GetHashCode();
+                hash = hash * 59 + this.ScreenWidth.GetHashCode();
+                hash = hash * 59 + this.TimeZoneOffset.GetHashCode();
+                hash = hash * 59 + this.JavaEnabled.GetHashCode();
```
KadoBOT, line 206: I wonder if it makes sense to have a function where you pass the `this.ColorDepth.GetHashCode()` and the ones below, to calculate the new value, since the `hash = hash * 59 + ` part is being repeated a couple of times/

Adyen/Model/Checkout/DefaultPaymentMethodDetails.cs
```diff
@@ -271,5 +275,69 @@ IEnumerable<System.ComponentModel.DataAnnotations.ValidationResult> IValidatable
...
+        /// Gets the hash code
+        /// </summary>
+        /// <returns>Hash code</returns>
+        public override int GetHashCode()
+        {
+            unchecked // Overflow is fine, just wrap
+            {
+                int hashCode = 41;
+                if (this.Type != null)
+                    hashCode = hashCode * 59 + this.Type.GetHashCode();
```
KadoBOT, line 290: the same comment as before. If someday you need to change this, you change just in one place. I.e.: changing the `59` to `60`

Adyen/Model/Checkout/DefaultPaymentMethodDetails.cs
```diff
@@ -102,7 +102,9 @@ public override string ToString()
             sb.Append("  EncryptedExpiryMonth: ").Append(EncryptedExpiryMonth).Append("\n");
             sb.Append("  EncryptedExpiryYear: ").Append(EncryptedExpiryYear).Append("\n");
             sb.Append("  EncryptedSecurityCode: ").Append(EncryptedSecurityCode).Append("\n");
+#pragma warning disable 618
             sb.Append("  RecurringDetailReference: ").Append(RecurringDetailReference).Append("\n");
+#pragma warning restore 618
```
msilvagarcia, line 107: Would be nice to add a comment of what's being done here

Adyen/Model/BrowserInfo.cs
AlexandrosMor: This code was autogenerated from swagger using Adyen open api specs for checkout. If I do use a method for it (which is nicer) I have to do the same for the rest models and I think it is outside of the ticket scope.

Adyen/Model/Checkout/DefaultPaymentMethodDetails.cs
AlexandrosMor: Check my comment above :)

Adyen/Model/Checkout/DefaultPaymentMethodDetails.cs
AlexandrosMor: There are some fields that are obsolete and raise warnings if are used. We are still using them for backward compatibility as removing them will break implementations. I decided to suppress the warning as we know the cause.    [Here ](https://stackoverflow.com/questions/968293/c-sharp-selectively-suppress-custom-obsolete-warnings) there is a guide how to suppress warnings in c#

PR #322, opened 2020-09-21
Adding New Property to BalanceInquiryResponse

SN185205:
we are excepting similar behavior as Store Value response for getting Payment Receipt in BalanceInquiryResponse. This will help us to get gift card balance enquiry receipt. Currently we are not sure whether this will work, but would like give a try. please let us know you are thoughts.

## Tested scenarios

**Fixed issue**:


AlexandrosMor (approved):

AlexandrosMor:
Hello @SN185205 
Thank you for the pull request, We see that indeed this property should be part of BalanceInquiryResponse as it is set in the [documentation](https://docs.adyen.com/point-of-sale/terminal-api-reference#comadyennexobalanceinquiryresponse)  

regards,
Alexandros

msilvagarcia (approved):

Issue #330, opened 2020-10-01
DefaultPaymentMethodDetails does not support new version of apple pay


Melchy:
**Describe the bug**
Adyen api changed /payments enpoint in version 49 for Apple pay. Property additionalData.applepay.token was renamed to applepay.token. 

Object [DefaultPaymentMethodDetails](https://github.com/Adyen/adyen-dotnet-api-library/blob/develop/Adyen/Model/Checkout/DefaultPaymentMethodDetails.cs) is incompatible with this change. It still uses old name for ApplePayToken.

**Expected behavior**
Class DefaultPaymentMethodDetails should contain property ApplePayToken with attribute `[DataMember(EmitDefaultValue = false, Name = "applepay.token")]`

Is this behavior intentional? Are there any plans to rename this property? (Adyen apple pay SDK is currently incompatible with this object.)

AlexandrosMor:
Hello @Melchy 

We are currently working to support Adyen api version v64. This fix will be included too. We hope that we will be able to pick it up next sprint

kind regards,
Alexandros

AlexandrosMor:
Hello @Melchy,

This is fixed from this pr #338. It will be added with the next release 

 kind regards,
Alexandros

PR #338, opened 2020-10-30
Corrects the property name for Apple Pay tokens

jimmah:
## Summary
We currently parse submitted payments from the front-end in our system using `Adyen.Model.Checkout.DefaultPaymentMethodDetails`, but ran into an issue where Apple Pay tokens weren't present, resulting in the following error when POST-ing the payment to Adyen:

    {
       "status": 422,
       "errorCode": "14_004",
       "message": "Missing payment method details",
       "errorType: "validation"
    }

After investigating and confirming that the frontend was indeed sending `applepay.token` as expected, I noticed issue #330 which is the exact problem we have. Comparing with the Adyen Java library, the same issue isn't present.

## Tested scenarios
Apple Pay payment was successful using Adyen.dll built from this change.

**Fixed issue**:  #330

### DIFF
--- a/Adyen.Test/ApplePayDeserializationTest.cs
+++ b/Adyen.Test/ApplePayDeserializationTest.cs
@@ -1,0 +1,23 @@
+﻿using Adyen.Model.Checkout;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
+
+namespace Adyen.Test
+{
+    [TestClass]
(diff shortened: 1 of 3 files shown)


AlexandrosMor:
Hello @jimmah,
 
Thank you for opening this PR. Could also add a unit test regarding this change?

kind regards,
Alexandros

AlexandrosMor (approved):

msilvagarcia (approved):

AlexandrosMor:
fixes #330

PR #347, opened 2020-11-26
[PW-2640] Deserialize the response of dropin into the correct paymentMethod model

AlexandrosMor:
## Summary
In this pr we are extending the Newtonsoft JsonConverter with a custom converter to deserialize the payment request with the payment method details class based on the IPaymentMethodDetails interface. This deserializes the paymentmethod based on the type and return the corresponding payment method detail class instance.
## Tested scenarios 

Test the deserialization with payment request and payment method detail class

**Fixed issue**:

PR #354, opened 2020-12-09
Added missing SubType property and unit test for it.

NemanjaMilenkovic990:
## Summary

Describe the changes proposed in this pull request:
- We noticed that SubType property is missing from the DefaultPaymentMethodDetails.cs and because of that PayPay smart payment was not able to trigger. 

When this property is added PayPay smart option becomes available for customers.

## Tested scenarios

Added Unittest that covers if SubType propery is present and if its value is the one expected for PayPal smart option.

**Fixed issue**:

### DIFF
--- a/Adyen.Test/BaseTest.cs
+++ b/Adyen.Test/BaseTest.cs
@@ -152,6 +152,23 @@ namespace Adyen.Test
             return paymentsRequest;
         }
 
+        public Model.Checkout.PaymentRequest CreatePayPalSmartPaymentRequestCheckout()
+        {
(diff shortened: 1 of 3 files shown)


AlexandrosMor:
Hello @NemanjaMilenkovic990,

This issue is solved with this [PR](https://github.com/Adyen/adyen-dotnet-api-library/pull/347) and I added a unit test for paypal [here](https://github.com/Adyen/adyen-dotnet-api-library/pull/355) 

Could you check if these solve the subtype issue for you ? 

Kind regards,
Alexandros

AlexandrosMor:
This issue is fixed in the latest release 6.0.0

PR #369, opened 2021-01-19
Add serializers to model

deepu105:
## Summary
The new custom deserializer are nice but would be nicer if they are declared in the Model object as well so that they can be deserialized directly in the request body in ASP.NET

## Tested scenarios
Tests added

**Fixed issue**:


AlexandrosMor (approved):

deepu105:
@msilvagarcia can you take a look at this plz, would be nice to have it in the next release for completion

msilvagarcia (approved):

PR #419, opened 2021-04-02
[PW-4329] Update to Adyen api v67

AlexandrosMor:
## Summary

Update the library to support api version 67. The `action` is not part of the payments/details response. There is a seperate model  `PaymentDetailsResponse`

## Tested scenarios

Tested with the mock and end to end tests

**Fixed issue**:

PR #416, opened 2021-03-29
Add DisableAsync to Recurring

johanssontobias:
## Summary
This PR adds an asynchronous method to disable recurring contracts. DisableAsync in Recurring.

## Tested scenarios
Uses an equivalent test as the synchronous Disable method.

PR #422, opened 2021-04-12
Release 7.0.0

### LINKED ISSUE #298: [PW-2421] Can you strong name the Adyen Nuget Package?
Hi,

My company uses strong named dlls for all of our C# projects. As such we can only use 3rd party nuget packages "out of the box" if they are also strong named. The Adyen nuget package is not strong named and therefore we have to do some additional steps to work around that.

Because it is not strong named we either have to
1. Compile from source and sign it during compilation (what we are doing now) or,
2. Disassemble your dll to IL and then rebuild and sign it (see https://stackoverflow.com/questions/331520/how-to-fix-referenced-assembly-does-not-have-a-strong-name-error)

Neither of these is ideal and both make it hard to upgarde to newer versions of your SDK. Both could be solved by Adyen strong naming their assemblies

**Describe the solution you'd like**
I would like for Adyen to strong name their assemblies.

AlexandrosMor:
## Summary

#419 [PW-4329] Update to Adyen api v67 
#416 Add DisableAsync to Recurring 
#414 [PW-4400] Add the missing fields for BacsDirectDebit 
#405 [PW-2421] Strong-name assembly 
#386 [PW-4118]Add the same exception handling for request and request async 
#383 PaymentLinksAsync 

## Tested scenarios

Mock tests
End to end tests
**Fixed issue**:  
fixes #298


msilvagarcia (approved):

Aleffio (dismissed):

msilvagarcia (dismissed):

msilvagarcia (approved):

deepu105 (dismissed):

msilvagarcia (dismissed):

Adyen/Service/Checkout.cs
```diff
@@ -106,11 +106,11 @@ public async Task<PaymentMethodsResponse> PaymentMethodsAsync(PaymentMethodsRequ
         /// </summary>
         /// <param name="paymentsDetailsRequest"></param>
         /// <returns>PaymentsResponse</returns>
```
Aleffio, line 108:
```suggestion
        /// <returns>PaymentDetailsResponse</returns>
```

Adyen/Service/Checkout.cs
```diff
@@ -106,11 +106,11 @@ public async Task<PaymentMethodsResponse> PaymentMethodsAsync(PaymentMethodsRequ
...
-        public PaymentResponse PaymentDetails(PaymentsDetailsRequest paymentsDetailsRequest, RequestOptions requestOptions = null)
+        public PaymentDetailsResponse PaymentDetails(PaymentsDetailsRequest paymentsDetailsRequest, RequestOptions requestOptions = null)
         {
             var jsonRequest = Util.JsonOperation.SerializeRequest(paymentsDetailsRequest);
             var jsonResponse = _paymentDetails.Request(jsonRequest, requestOptions);
-            return Util.JsonOperation.Deserialize<PaymentResponse>(jsonResponse);
+            return Util.JsonOperation.Deserialize<PaymentDetailsResponse>(jsonResponse);
         }
 
         /// <summary>
```
Aleffio, line 116: should the `PaymentDetailsAsync` method return a  `Task<PaymentDetailsResponse>` instead?

Adyen/Service/Checkout.cs
```diff
@@ -184,5 +184,17 @@ public PaymentLinkResource PaymentLinks(CreatePaymentLinkRequest createPaymentLi
             var jsonResponse = _paymentLinksResult.Request(jsonRequest);
             return JsonConvert.DeserializeObject<PaymentLinkResource>(jsonResponse);
         }
+
+        /// <summary>
+        /// POST /paymentsLinks API call async
+        /// </summary>
+        /// <param name="createPaymentLinkRequest"></param>
+        /// <returns>CreatePaymentLinkResponse</returns>
```
Aleffio, line 192: minor
```suggestion
        /// <returns>PaymentLinkResource</returns>
```
not sure if `Task` should be included too

Aleffio (approved):

msilvagarcia (approved):

Issue #288, opened 2020-06-27
[PW-2616] SaleToPoiMessageSecuredEncryptor.Encrypt uses ASCII encoding instead of UTF-8


bojan-kukricar:
**Description**
During encryption a message is ASCII encoded which results in an issue for example during printing.
Characters which do not exist in ASCII table, will not be correctly interpreted.

      public SaleToPoiMessageSecured Encrypt(string saleToPoiMessage, MessageHeader messageHeader,
                                                 EncryptionCredentialDetails encryptionCredentialDetails)
        {
            var encryptionDerivedKey = _encryptionDerivedKeyGenerator.Generate(encryptionCredentialDetails);
            var saleToPoiMessageJson = saleToPoiMessage;
            var saleToPoiMessageByteArray = Encoding.ASCII.GetBytes(saleToPoiMessageJson);
            var ivMod = _ivModGenerator.GenerateRandomMod();
            var saleToPoiMessageAesEncrypted = _aesEncryptor.Encrypt(saleToPoiMessageByteArray,
                                                                     encryptionDerivedKey,
                                                                     ivMod);
            var saleToPoiMessageAesEncryptedHmac = _hmacSha256Wrapper.HMac(saleToPoiMessageByteArray,
                                                                           encryptionDerivedKey.HmacKey);

            var saleToPoiMessageSecured = new SaleToPoiRequestSecured
            {
                MessageHeader = messageHeader,
                NexoBlob = Convert.ToBase64String(saleToPoiMessageAesEncrypted),
                SecurityTrailer = new SecurityTrailer
                {
                    KeyVersion = encryptionCredentialDetails.KeyVersion,
                    KeyIdentifier = encryptionCredentialDetails.KeyIdentifier,
                    Hmac = saleToPoiMessageAesEncryptedHmac,
                    Nonce = ivMod,
                    AdyenCryptoVersion = encryptionCredentialDetails.AdyenCryptoVersion
                }
            };
            
            return saleToPoiMessageSecured;
        }

**To Reproduce**
1. Create a print request with OpenText.Text="m\u006DÄ\u00C4"
2. Execute
3. Terminal will print "mm??"

AlexandrosMor:
hello @bojan-kukricar,

We are going to create an internal ticket and investigate this issue.

kind regards,
Alexandros

AlexandrosMor:
Hello @bojan-kukricar,

Could you send us a PrintRequest sample so we can test it on our end? 

[...]

PR #423, opened 2021-04-12
Changing encoding from ASCII to UTF8 to account for special character…

Katie-Shields:
…s in print request for terminals

## Summary
Merchants were reporting issue with special characters when trying to use the printRequest on the POS terminal. Specifically they were testing with german characters, they were printed as question marks on the receipt. This is because these characters don't exist in ASCII so switched to UTF8.

## Tested scenarios
Tested with merchant test app, on my V400m. after changing to UTF8 there were no issues printing.

**Fixed issue**:

### DIFF
--- a/Adyen.Test/MockPosApiRequest.cs
+++ b/Adyen.Test/MockPosApiRequest.cs
@@ -110,14 +110,14 @@ namespace Adyen.Test
         }
 
         /// <summary>
-        /// Dummy Nexo json print request 
+        /// Dummy Nexo json print request
         /// </summary>
         /// <returns></returns>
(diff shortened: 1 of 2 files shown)


AlexandrosMor:
Hello @Katie-Shields, 

Thank you for spotting this bug and opening a pull request. Could you please correct the mock test that is failing ?

kind regards, 
Alexandros
Adyen

AlexandrosMor (approved):

msilvagarcia (approved):

bojan-kukricar:
This is what I reported here #288. When do you plan the next release?

PR #438, opened 2021-05-17
[PW-4465] Add new endpoint /notifyShopper

AlexandrosMor:
## Summary

As part of the India go LIVE, there are some fields needed for integration. This will include the mandate field in the request body for Checkout API, and the introduction of a new notifyShopper endpoint.

This document includes the latest changes:
https://docs.adyen.com/payment-methods/cards/cards-recurring-india
## Tested scenarios

**Fixed issue**:


Adyen/Constants/ClientConfig.cs
```diff
@@ -41,7 +41,7 @@ public class ClientConfig
         public static string MarketPayFundApiVersion = "v5";
         public static string MarketPayAccountApiVersion = "v5";
         public static string HopApiVersion = "v1";
-        public static string RecurringApiVersion = "v25";
+        public static string RecurringApiVersion = "v49";
```
deepu105, line 44: wow thats a huge bumb, is there only this many change in model? I would expect more model changes from such a bump

deepu105 (commented):
Looks fine, but for the recurring API version bumb, is this the only change in models and API?

Adyen/Constants/ClientConfig.cs
AlexandrosMor: Right, I updated the models based on the open api specs for v49

deepu105 (approved):
looks good

Aleffio (approved):

PR #441, opened 2021-05-25
Payment completion details: ReturnUrlQueryString

mattias-ancon:
Added ReturnUrlQueryString to support swish payments

## Summary

This changes Adyen/Model/Checkout/PaymentCompletionDetails.cs to include the property **ReturnUrlQueryString** 

**- What is the motivation for this change?** 
We needed support for swish payments and the ios version of this sdk https://github.com/Adyen/adyen-ios/releases/tag/4.0.0-beta.2 will send that property but it was not handled in this server sdk.

## Tested scenarios
Tested payments with swish from the ios sdk 

https://user-images.githubusercontent.com/84770613/119466356-d130e380-bd44-11eb-9617-d47989da8851.mp4

**Fixed issue**:

### DIFF
--- a/Adyen/Model/Checkout/PaymentCompletionDetails.cs
+++ b/Adyen/Model/Checkout/PaymentCompletionDetails.cs
@@ -63,7 +63,7 @@ namespace Adyen.Model.Checkout
             string payerID = default(string), string payload = default(string), string paymentID = default(string),
             string paymentStatus = default(string), string redirectResult = default(string),
             string threeDSResult = default(string), string threeds2ChallengeResult = default(string),
-            string threeds2Fingerprint = default(string))
+            string threeds2Fingerprint = default(string), string returnUrlQueryString = default(string))
         {
             this.MD = mD;
             this.PaReq = paReq;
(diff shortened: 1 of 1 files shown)


deepu105 (approved):
LGTM

deepu105 (commented):
Does this need an API version update @AlexandrosMor ?

mattias-ancon:
Hi all i also request a support ticket asking about this.

Perhaps this PR is unneeded if i can map to redirectResult instead?
What do you think?

This was the reply:
----------------------------------------------------------
Tony Star (Adyen)
May 26, 2021, 8:43 GMT+2
Hi Mattias,

returnUrlQueryString was a field to be submitted in the /payments/details request 3 years ago, e.g Dotpay requires this field back to 2018.

As we want all LPMs(local payment methods) to align with each other in /payments/details request, our developer replaced returnUrlQueryString with payload, and then redirectResult (in v66 and later).

Our documentation team therefore marked this field as hidden in the API explorer page.

Let us know if you need any further assistance.

AlexandrosMor (approved):

PR #476, opened 2021-07-08
Add more constants for additional data 

### LINKED ISSUE #420: Add more constants for additional data
Currently we have these constants available, which is useful when retrieving data from webhook notification, e.g. to verify `HmacSignature`: https://github.com/Adyen/adyen-dotnet-api-library/blob/master/Adyen/Constants/AdditionalData.cs

However I noticed a few addtional data returned from webhook notifications, which are not included: "paymentLinkId" and "merchantOrderReference".

For now I have added these in my own constant class plus some custom meta data prefixed with `metadata.`.

```
public class Constants
{
    // Adyen constants in .NET library: https://github.com/Adyen/adyen-dotnet-api-library/blob/master/Adyen/Constants/AdditionalData.cs
    
    public class AdditionalData
    {
        // Additional constants not included in library
        public const string PaymentLinkId = "paymentLinkId";
        public const string MerchantOrderReference = "merchantOrderReference";

        // Custom constants for metadata
        public const string OrderReference = "metadata.orderReference";
        public const string OrderId = "metadata.orderId";
        public const string OrderNumber = "metadata.orderNumber";
    }
}
```

Unfortunately the CANCELLATION notification currently only return `hmacSignature`, even though the "meta data" setting in enabled under addtional settings for the Standard Notification in Adyen. I have reached out to the support, but unfortunately this doesn't seem to have a high prioritization.

AlexandrosMor:
## Tested scenarios

**Fixed issue**:  
fixes #420


Adyen/Constants/AdditionalData.cs
```diff
@@ -46,5 +46,11 @@ public class AdditionalData
         public const string RecurringShopperReference = "recurring.shopperReference";
         public const string RecurringDetailReference = "recurring.recurringDetailReference";
         public const string AllowThreeDS2 = "allow3DS2";
+        public const string PaymentLinkId = "paymentLinkId";
+        public const string MerchantOrderReference = "merchantOrderReference";
+        // Custom constants for metadata
+        public const string OrderReference = "metadata.orderReference";
+        public const string OrderId = "metadata.orderId";
+        public const string OrderNumber = "metadata.orderNumber";
```
bjarnef, line 54: @AlexandrosMor `OrderReference`, `OrderId` and `OrderNumber` should probably not be included. These were just custom meta data properties I sent in the request from the ecommerce system, but not something specific in Adyen.

AlexandrosMor: Thank you for letting us know :)

Adyen/Constants/AdditionalData.cs
```diff
@@ -46,5 +46,11 @@ public class AdditionalData
         public const string RecurringShopperReference = "recurring.shopperReference";
         public const string RecurringDetailReference = "recurring.recurringDetailReference";
         public const string AllowThreeDS2 = "allow3DS2";
+        public const string PaymentLinkId = "paymentLinkId";
+        public const string MerchantOrderReference = "merchantOrderReference";
+        // Custom constants for metadata
+        public const string OrderReference = "metadata.orderReference";
```
AlexandrosMor, line 52:
```suggestion

```

Adyen/Constants/AdditionalData.cs
```diff
@@ -46,5 +46,11 @@ public class AdditionalData
         public const string RecurringShopperReference = "recurring.shopperReference";
         public const string RecurringDetailReference = "recurring.recurringDetailReference";
         public const string AllowThreeDS2 = "allow3DS2";
+        public const string PaymentLinkId = "paymentLinkId";
+        public const string MerchantOrderReference = "merchantOrderReference";
+        // Custom constants for metadata
+        public const string OrderReference = "metadata.orderReference";
+        public const string OrderId = "metadata.orderId";
```
AlexandrosMor, line 53:
```suggestion
  

```

Adyen/Constants/AdditionalData.cs
AlexandrosMor:
```suggestion

```

Adyen/Constants/AdditionalData.cs
bjarnef: I think `OrderNumber` should be removed as well as Adyen doesn't return this meta data and this was a custom meta data I sent in my use-case.

As I recall the `reference` property in the request is returned as `merchantOrderReference` in webhook notification.

Aleffio (approved):

deepu105 (approved):

Issue #475, opened 2021-07-06
Consider migrating from HttpWebRequest to HttpClient


RyanONeill1970:
When trying to debug the dot net library we are unable to view the communications between Adyen and the client server because HttpWebRequest does not have logging built in. Exceptions are also thrown in HttpWebRequest and have to be caught and parsed to get the status code in some situations. This last part is noisy (shows in logs even if handled), slow and error prone.

The new HttpClient approach;

- Can be logged easily via the appsettings, enabling traffic to be monitored for debugging.
- Handles connection pooling via the HttpClientFactory class, meaning better use of sockets which helps in a busy app.
- Can be swapped out for testing via the IHttpClient interface.
- Does not throw exceptions for standard status codes such as 404 instead returns the status with an IsSuccessStatusCode property.
- Is .Net Standard compatible and is the now standard approach.
- Can still be used from older .net versions.

I appreciate it is a framework level thing and not related to the product, it would help us a lot whilst developing and debugging.

AlexandrosMor:
Hello @RyanONeill1970,

It is a decision from our side to not add any logging into the libraries. On the other hand you are more than welcome to implement the suggestions which we could review them. You can follow the [contributing guidelines](https://github.com/Adyen/adyen-dotnet-api-library/blob/develop/CONTRIBUTING.md) 

kind regards,
Alexandros
Adyen

kipwoker:
@AlexandrosMor it's only about the logging. The current solution has performance issues. Under the hood `httpWebRequest.GetResponse()` method creates HttpClient for each call.

Take a look at this article for more information: https://josef.codes/you-are-probably-still-using-httpclient-wrong-and-it-is-destabilizing-your-software/

jillingk:
Hi @RyanONeill1970 and @kipwoker,

Thank you both for reaching out to us and contributing. We have finally arrived at this library with our automation process and will thus start very soon with some big changes. This includes the HttpClient, as it will need to be RESTful anyways. We will be sure to use your feedback during this process. If there is anything you guys would like to see, please do let us know.

Best, Jilling
Adyen

PR #477, opened 2021-07-12
Allow to use HttpClient instead of HttpWebRequest

kipwoker:
## Summary
HttpWebRequest produces performance issues.

Take a look at these articles for more information:
- https://josef.codes/you-are-probably-still-using-httpclient-wrong-and-it-is-destabilizing-your-software/
- https://makolyte.com/csharp-switch-from-using-httpwebrequest-to-httpclient/
- https://www.aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/

Proposed changes:
- Allow passing user's `HttpClient` that could be created through the `HttpClientFactory` for ASP.NET Core apps or manually for others
- Rename `HttpUrlConnectionClient` to `HttpWebRequestWrapper` and use this implementation for backward compatibility

**Fixed issue**:  #475


kipwoker:
@AlexandrosMor almost a month passed. Can anybody take a look at the PR, please? :)

AlexandrosMor:
@kipwoker We are currently busy as we maintain all the repos in the Adyen's github page. This pr needs extra attention as modifies core functionality. It is in our todo list :)

Adyen/HttpClient/HttpClientExtensions.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.Net.Http;
```
AlexandrosMor, line 1:
```suggestion
#region License
// /*
//  *                       ######
//  *                       ######
//  * ############    ####( ######  #####. ######  ############   ############
//  * #############  #####( ######  #####. ######  #############  #############
//  *        ######  #####( ######  #####. ######  #####  ######  #####  ######
//  * ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
//  * ###### ######  #####( ######  #####. ######  #####          #####  ######
//  * #############  #############  #############  #############  #####  ######
//  *  ############   ############  #############   ############  #####  ######
//  *                                      ######
//  *                               #############
//  *                               ############
//  *
//  * Adyen Dotnet API Library
//  *
//  * Copyright (c) 2021 Adyen B.V.
//  * This file is open source and available under the MIT license.
//  * See the LICENSE file for more info.
//  */
#endregion
﻿using System.Net.Http;
```

Adyen.Test/HttpClientWrapperTest.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
```
AlexandrosMor, line 1: is this class used ?

Adyen.Test/HeaderRequestTest.cs
```diff
@@ -30,18 +30,17 @@ namespace Adyen.Test
     [TestClass]
     public class HeaderRequestTest
     {
-
-        private readonly HttpUrlConnectionClient _httpUrlConnectionClient;
-        private readonly string _endpoint = "https://endpoint:8080";
+        private const string _endpoint = "https://endpoint:8080";
 
         public HeaderRequestTest()
         {
```
AlexandrosMor, line 36: If this constructor is not used we can remove it

Adyen/HttpClient/HttpWebRequestWrapper.cs
```diff
@@ -223,5 +206,9 @@ public static string QueryString(IDictionary<string, string> dict)
             }
             return string.Join("&", list);
         }
+
+        public void Dispose()
```
AlexandrosMor, line 210: Is this needed?

kipwoker: It's needed for backward compatibility only since the new implementation uses `IDisposable` pattern and the users of `IClient` interface shouldn't take care of that (just implemented encapsulation principle).

Adyen.Test/HeaderRequestTest.cs
kipwoker: Good point. Will remove

Adyen.Test/HttpClientWrapperTest.cs
kipwoker: Sorry, I think I just forgot to delete this one. Made it for some experiments.

Adyen/HttpClient/HttpWebRequestWrapper.cs
kipwoker: The main idea was about giving backward compatibility for library users. So if they don't use HttpClient yet everything should work more or less the same way. With this approach, the version upgrade should be softer, without breaking changes. :)

Adyen.Test/HttpClientWrapperTest.cs
kipwoker: Removed

Adyen.Test/HeaderRequestTest.cs
kipwoker: Removed

Adyen/HttpClient/HttpWebRequestWrapper.cs
```diff
@@ -167,53 +162,41 @@ public string Post(string endpoint, Dictionary<string, string> postParameters, C
             return new StreamReader(response.GetResponseStream()).ReadToEnd();
         }
 
-        public HttpWebRequest GetHttpWebRequest(string endpoint, Config config, bool isApiKeyRequired, RequestOptions requestOptions = null)
+        public HttpWebRequest GetHttpWebRequest(string endpoint, bool isApiKeyRequired, RequestOptions requestOptions = null)
         {
             //Add default headers
             var httpWebRequest = (HttpWebRequest)WebRequest.Create(endpoint);
```
AlexandrosMor, line 168: The purpose of this PR was to remove the `HttpWebRequest`

Adyen/HttpClient/HttpClientWrapper.cs
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        public async Task<string> RequestAsync(string endpoint, string json, bool isApiKeyRequired, RequestOptions requestOptions = null)
+        {
+            using (var request = GetHttpRequestMessage(endpoint, isApiKeyRequired, json, requestOptions))
+            using (var httpResponseMessage = await _httpClient.SendAsync(request))
+            {
+                var responseText = await httpResponseMessage.Content.ReadAsStringAsync();
+                httpResponseMessage.EnsureSuccessStatusCode();
+                return responseText;
```
AlexandrosMor, line 67: There is no exception handling here. Is there any reason ?

Adyen/HttpClient/HttpClientWrapper.cs
```diff
@@ -0,0 +1,140 @@
...
+        public string Post(string endpoint, Dictionary<string, string> postParameters)
+        {
+            var dictToString = QueryString(postParameters);
+
+            var content = new StringContent(dictToString, _encoding, "application/x-www-form-urlencoded");
+            using (var request = new HttpRequestMessage(HttpMethod.Post, endpoint) {Content = content})
+            using (var response = _httpClient.SendAsync(request).GetAwaiter().GetResult())
+            {
+                response.EnsureSuccessStatusCode();
+                return response.Content.ReadAsStringAsync().GetAwaiter().GetResult();
```
AlexandrosMor, line 80: Similar to here

Adyen/HttpClient/HttpWebRequestWrapper.cs
kipwoker: I noted this in the description
> Rename HttpUrlConnectionClient to HttpWebRequestWrapper and use this implementation for backward compatibility

I left this for backward compatibility, so the users of the library can use the old approach if they are stick to HttpWebRequest.

Adyen/HttpClient/HttpClientWrapper.cs
kipwoker: `EnsureSuccessStatusCode` throwing an exception that could be handled by the user if the status code is not succesful. What kind of handling are you expecting here?

Adyen/Client.cs
```diff
@@ -1,163 +1,180 @@
...
+            Config = new Config
+            {
+                Environment = environment,
+                XApiKey = xapikey
+            };
+            _httpClient = httpClient;
+            this.SetEnvironment(environment, liveEndpointUrlPrefix);
+        }
+
+        public Client(Config config, System.Net.Http.HttpClient httpClient = null)
```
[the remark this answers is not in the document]
kipwoker: Fixed

Adyen/Config.cs
```diff
@@ -51,9 +51,12 @@ public class Config
...
         [Obsolete("This is deprecated property by Adyen.")]
         public bool SkipCertValidation { get; set; } = false;
-      
-        //POS Terminal Management 
+
+        //POS Terminal Management
         public string PosTerminalManagementEndpoint { get; set; }
 
+
+        public bool HasPassword => !string.IsNullOrEmpty(Password);
```
AlexandrosMor, line 59: could we use getters for these
```suggestion
       public bool HasPassword
        {
            get { return !string.IsNullOrEmpty(Password); }
        }

```

Adyen/Config.cs
```diff
@@ -51,9 +51,12 @@ public class Config
...
+        public bool HasApiKey => !string.IsNullOrEmpty(XApiKey);
```
AlexandrosMor, line 60:
```suggestion
        public bool HasApiKey
        {
            get { return !string.IsNullOrEmpty(XApiKey); }
        }

```

Adyen/HttpClient/HttpClientWrapper.cs
```diff
@@ -0,0 +1,140 @@
...
+//  * ###### ######  #####( ######  #####. ######  #####          #####  ######
+//  * #############  #############  #############  #############  #####  ######
+//  *  ############   ############  #############   ############  #####  ######
+//  *                                      ######
+//  *                               #############
+//  *                               ############
+//  *
+//  * Adyen Dotnet API Library
+//  *
+//  * Copyright (c) 2020 Adyen B.V.
```
AlexandrosMor, line 19:
```suggestion
//  * Copyright (c) 2021 Adyen B.V.
```

Adyen/HttpClient/HttpClientWrapper.cs
```diff
@@ -0,0 +1,140 @@
...
+                httpWebRequest.Headers.Add("x-api-key", _config.XApiKey);
+            }
+            else if (_config.HasPassword)
+            {
+                var authString = _config.Username + ":" + _config.Password;
+                var bytes = Encoding.UTF8.GetBytes(authString);
+                var credentials = Convert.ToBase64String(bytes);
+                httpWebRequest.Headers.Add("Authorization", "Basic " + credentials);
+            }
+
```
AlexandrosMor, line 111: I think here we can add the certificate validation for the local terminal API for POS 
similar to the [httpwebrequest](https://github.com/Adyen/adyen-dotnet-api-library/blob/develop/Adyen/HttpClient/HttpURLConnectionClient.cs#L200)
The validation is used to verify the certificate on the terminal against the public key.

Adyen/HttpClient/HttpClientWrapper.cs
AlexandrosMor: Indeed. We changed the exception handling on the [old HttpURLConnectionClient](https://github.com/Adyen/adyen-dotnet-api-library/blob/develop/Adyen/HttpClient/HttpURLConnectionClient.cs#L118). I was thinking if the `HandleWebException` is outsourced in another class if could we use the same here.

Adyen/Config.cs
kipwoker: Well, of course, we could, but it has no sense because it's just a syntax sugar that was introduced to reduce the boilerplate

Adyen/HttpClient/HttpClientWrapper.cs
kipwoker: Actually, it's impossible to do it there. For `HttpClient` you can set up it through the `HttpClientHandler` during initialization only. It's impossible to configure it for a particular request. That's why I encapsulated this logic at `HttpClientExtensions.ConfigureHttpMessageHandler` method.

AlexandrosMor: Indeed but the `HttpClientExtensions.ConfigureHttpMessageHandler` is not used by the `HttpClient`. Could you provide us an example where to set up it through the `HttpClientHandler` during initialisation?

jonekdahl:
This seems like an important improvement of the library. Is there anything we can do to help get it merged (testing etc.)?

Roeland54:
@AlexandrosMor Would it be possible to pick this up again? We would really benefit from this change.

AlexandrosMor:
@Roeland54 Thank you for the message, the issue is that we are in the process to automate some parts of the library so we are putting some effort on that. Replacing the client and also support rest is in our plans but we are still busy. I hope in the coming month to be able to pick this one

AlexandrosMor:
We are currently working on the new http client in this pr #690

PR #523, opened 2021-11-15
[PW-5828]Add application info in payment request service

AlexandrosMor:
## Summary

The application info was part of payment request added manually. That means every time that we create the models the application info should be added again. Now we included in the service so it is not part of the payment request model.
## Tested scenarios

Update the unit tests

**Fixed issue**:

### DIFF
--- a/Adyen/Service/Checkout.cs
+++ b/Adyen/Service/Checkout.cs
@@ -23,6 +23,7 @@
 
 using System.Threading.Tasks;
 using Adyen.Model;
+using Adyen.Model.ApplicationInformation;
 using Adyen.Model.Checkout;
 using Adyen.Service.Resource.Checkout;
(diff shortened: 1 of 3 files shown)


Adyen/Service/Checkout.cs
```diff
@@ -67,6 +68,12 @@ public Checkout(Client client) : base(client)
         /// <returns>PaymentsResponse</returns>
         public PaymentResponse Payments(PaymentRequest paymentRequest, RequestOptions requestOptions = null)
         {
+            //if it is set by the merchant do not override
+            if (paymentRequest.ApplicationInfo == null)
```
wboereboom, line 72: Do we want to allow users to change the application? This might skew any data we collect on library use. What are the use cases?

PR #529, opened 2021-11-30
[PW-5676] Update MarketPay/Transaction.cs TransactionStatusEnum

### LINKED ISSUE #506: [PW-5676] Fee is missing in Adyen.Model.MarketPay.Transaction.TransactionStatusEnum
Hello

After upgrading to version 8 i got an error when calling FundService and method AccountHolderTransactionList.

Error message:
`Error converting value "Fee" to type System.Nullable1[Adyen.Model.MarketPay.Transaction+TransactionStatusEnum]. Path accountTransactionLists[0].transactions[12].transactionStatus, line 1, position 5501.`

AlexandrosMor:
## Tested scenarios

Update TransactionStatusEnum based on the [open api specs FundService-v5](https://github.com/Adyen/adyen-openapi/blob/main/json/FundService-v5.json#L936)

**Fixed issue**:  
fixes #506

### DIFF
--- a/Adyen/Model/MarketPay/Transaction.cs
+++ b/Adyen/Model/MarketPay/Transaction.cs
@@ -33,152 +33,276 @@ namespace Adyen.Model.MarketPay
     /// <summary>
     /// Transaction
     /// </summary>
-    [DataContract]
-    public class Transaction : IEquatable<Transaction>, IValidatableObject
+    [DataContract(Name = "Transaction")]
+    public partial class Transaction : IEquatable<Transaction>, IValidatableObject
     {
         /// <summary>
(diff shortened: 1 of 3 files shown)


wboereboom (approved):
Hard to check for correctness. do we have test transactions for this change?

Adyen/Model/MarketPay/Transaction.cs
```diff
@@ -448,8 +585,7 @@ public bool Equals(Transaction input)
                 ) &&
                 (
                     this.TransactionStatus == input.TransactionStatus ||
-                    (this.TransactionStatus != null &&
-                    this.TransactionStatus.Equals(input.TransactionStatus))
+                    this.TransactionStatus.Equals(input.TransactionStatus)
```
zaiddreakh, line 588: is the `!= null` check not needed here? it seems to be used in all other statements

zaiddreakh (approved):

Adyen/Model/MarketPay/Transaction.cs
AlexandrosMor: An enum value cannot be null. It is a value type like an int.

Issue #387, opened 2021-02-15
[PW-4542] Add Async support for the Account Service


FrankVDL-Trivec:
For good practice all service call should be await able.

Can you add the xxxAsync implementation for all Service endpoints ?

AlexandrosMor:
Hello @FrankVDL-Trivec,

Thank you for the recommendation. We are going to create a ticket and pick it up as soon as possible. In the meantime feel free to open a pull request. 

kind regards,
Alexandros
Adyen

PR #608, opened 2022-05-24
Add Async support for the Account Service

akScript:
**Description**
Add Async support for the Account Service, this was done by adding an Async version of each method and calling:
```csharp
Adyen.Service.Resource.RequestAsync(string json)
```
This should hopefully fix issue #387 

**Tested scenarios**
Unit tests were added to `MarketpayAccountTest` for each Async method that was added.
All tests pass


AlexandrosMor (approved):

akScript:
@AlexandrosMor @rikterbeek, It seems that PRs require at least 2 approving reviewers, can you please look into this one, it will help a lot, thanks.

wboereboom (approved):

PR #626, opened 2022-07-07
[PW-6749]Add the certificate path for terminal api

AlexandrosMor:
**Description**

Merchant raised a security concern on how the Adyen CA root certificate is handled. They would like to avoid having to add the cert to the system store, as then it would provide too much access, they would much prefer to restrict it solely to communication with the terminals. For that reason we provide the possibility to pass it from certificate path and create a CA store using the `Certificate` class in `Adyen.Util`. This class should be used in application level when the merchant needs to create a ca store using the certificate path

**Tested scenarios**

Tested on windows and ubuntu with local communication. For windows it is always asking to install the certificate to the local CA store. It can be skipped on windows

### DIFF
--- a/Adyen/HttpClient/HttpURLConnectionClient.cs
+++ b/Adyen/HttpClient/HttpURLConnectionClient.cs
@@ -198,16 +198,6 @@ namespace Adyen.HttpClient
                 httpWebRequest.Proxy = config.Proxy;
             }
 
-            if (!string.IsNullOrEmpty(config.CertificatePath))
-            {
-                using (X509Store store = new X509Store(StoreName.Root, StoreLocation.CurrentUser))
-                {
-                    var x509Certificate2 = new X509Certificate2(config.CertificatePath);
-                    store.Open(OpenFlags.ReadWrite);
-                    store.Add(x509Certificate2);
-                }
-            }
-
             httpWebRequest.ServerCertificateValidationCallback = ServerCertificateValidationCallback;
             return httpWebRequest;
         }
--- a/Adyen/HttpClient/HttpURLConnectionClient.cs
+++ b/Adyen/HttpClient/HttpURLConnectionClient.cs
@@ -197,6 +197,17 @@ namespace Adyen.HttpClient
             {
                 httpWebRequest.Proxy = config.Proxy;
             }
+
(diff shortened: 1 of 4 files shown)


Adyen/HttpClient/HttpURLConnectionClient.cs
```diff
@@ -197,6 +197,17 @@ public HttpWebRequest GetHttpWebRequest(string endpoint, Config config, bool isA
...
                 httpWebRequest.Proxy = config.Proxy;
             }
+
+            if (!string.IsNullOrEmpty(config.CertificatePath))
+            {
+                using (X509Store store = new X509Store(StoreName.Root, StoreLocation.CurrentUser))
+                {
+                    var x509Certificate2 = new X509Certificate2(config.CertificatePath);
+                    store.Open(OpenFlags.ReadWrite);
+                    store.Add(x509Certificate2);
```
Kwok-he-Chu, line 207: I think we should have this code elsewhere. Several things to consider:

- Performance for opening/closing the store per HttpWebRequest call
- The GetHttpWebrequest-method should only be responsible for adding things to the HttpWebRequest itsself and not add certificates to a particular certificate store. What happens if this method is called again? config.CertificatePath will not be empty and we'll add the certificate again ;)

Kwok-he-Chu (commented):
Review

Adyen/HttpClient/HttpURLConnectionClient.cs
AlexandrosMor: I agree with the performance. The other calls works even we set a custom ca store with a specific certificate
Any ideas for the code location ?

Adyen/Util/Certificate.cs
```diff
@@ -0,0 +1,47 @@
+// #region License
```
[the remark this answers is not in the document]
AlexandrosMor: It is not in the plans for now I guess. At least I will not start  on this pr :)

Kwok-he-Chu (commented):
Fix your unittests > 'something related to TestAccount'(?) and it's approved! Great work :)

+1 Could you open a ticket which is about removing all these warnings (check annotations), I'll get rid of them when I have the time (saving you a ton of visual overload in the future during the automated builds or PR request).

Issue #585, opened 2022-04-12
[PW-6646] Add gift card support


andrewyakonyuk:
**Describe the bug**
At the moment it is not possible to complete the payment with gift card because of the error.

**To Reproduce**
I follow [Gift card component](https://docs.adyen.com/payment-methods/gift-cards/web-component) documentation to build my integration.
So, the drop-in component is used to collect the information from the user, provide it to backend and send it to `/payments` endpoint.
I use the following code to handle such cases (pseudocode):

```cs
// get methods from Adyen
var paymentMethods = new Checkout(client).PaymentMethods(methodsRequest);
// pass paymentMethods to drop-in component together with configuration
// user selects gift card option and provides all required information on the frontend
string componentData = null; // // collect drop-in component data as simple string

var request = JsonOperation.Deserialize<PaymentRequest>(componentData);
var response = new Checkout(client).Payments(request);
```

In the result the Adyen returns the following response:
> {"status":422,"errorCode":"158","message":"Required field 'brand' is not provided.","errorType":"validation","pspReference":"xxxx*5S82"}'

**Expected behavior**
Complete the payment with gift card without errors.

**Additional context**
The problem is that the PaymentRequest.PaymentMethod is deserialized into `DefaultPaymentMethodDetails` which does not have Brand property. So, it is good idea to add additional `GiftCardDetails` similar to `PaypalDetails` or 'GooglePayDetails'.

I created a PR to fix this issue: #586

wboereboom:
HI @andrewyakonyuk,

Thanks for raising this issue.
Are you able to serialize your response using [CardDetails](https://github.com/Adyen/adyen-dotnet-api-library/blob/develop/Adyen/Model/Checkout/CardDetails.c)? The `brand` field is in there, and the card type field can be set to [giftcard](https://github.com/Adyen/adyen-dotnet-api-library/blob/develop/Adyen/Model/Checkout/CardDetails.cs#L80)

Kind Regards,
Wouter
Adyen

AlexandrosMor:
Hello @andrewyakonyuk, 

Indeed there are some classes missing and we could generate them based on the [open api specs ](https://github.com/Adyen/adyen-openapi/blob/main/json/CheckoutService-v68.json). Adding classes which are generated would be removed at the next models generation. We created an internal ticket and we going to add those classes

kind regards,
Alexandros
Adyen

AlexandrosMor:
Hello @andrewyakonyuk,

[...]

PR #628, opened 2022-07-14
[PW-6646]Added brand field to DefaultPaymentMethodDetails

jillingk:
**Description**
With adding brand field to the class we can allow merchants to use giftcards. @AlexandrosMor, is this what you imagined with you proposed solution in the issue #585? 

**Tested scenarios**
Standard unit tests

**Fixed issue**:  #585


wboereboom (approved):

AlexandrosMor (approved):

huysentruitw:
Any idea when this change will be published to NuGet?

Issue #539, opened 2022-01-06
[PW-6739]Add properties to KYCVerificationResult: legalArrangements, legalArrangementsEntities, payoutMethods, signatories, ultimateParentCompany, verificationProfile


scottpantall:
**Is your feature request related to a problem? Please describe.**
The v6 `getAccountHolder` API endpoint (https://docs.adyen.com/api-explorer/#/Account/v6/post/checkAccountHolder) includes more KYCVerificationResult properties than the ones that exist in this library (https://github.com/Adyen/adyen-dotnet-api-library/blob/develop/Adyen/Model/MarketPay/KYCVerificationResult.cs).

**Describe the solution you'd like**
Add the legalArrangements, legalArrangementsEntities, payoutMethods, signatories, ultimateParentCompany, verificationProfile properties to the KYCVerificationResult and add all related classes for those properties.

jillingk:
Hi @scottpantall,

Thanks for reaching out to us. I created an internal ticket to add the properties to the library. If you would like to speed up this process consider[ contributing](https://github.com/Adyen/adyen-dotnet-api-library/blob/develop/CONTRIBUTING.md) and create a PR. 

Best, Jilling
Adyen

jillingk:
Hi @scottpantall,

Just a quick FYI that we updated the models! Apologies for the delay.

Best, Jilling
Adyen

PR #638, opened 2022-08-08
[PW-6739] Updated KYCVerificationResult Model and added additional

jillingk:
Updated KYCVerificationResult models and added additional dependent classes. 

**Description**
KYCVerificationResult was updated and added some other classes that the KYCVerificationResult was depended on. 

**Tested scenarios**
Ran unit tests.

**Fixed issue**: #539

### DIFF
--- a/Adyen/Model/MarketPay/KYCVerificationResult.cs
+++ b/Adyen/Model/MarketPay/KYCVerificationResult.cs
@@ -28,6 +28,7 @@ using System.IO;
 using System.Linq;
 using System.Runtime.Serialization;
 using System.Text;
+using Adyen.Util;
 using Newtonsoft.Json;
 
 namespace Adyen.Model.MarketPay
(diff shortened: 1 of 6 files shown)


Adyen/Model/MarketPay/KYCVerificationResult.cs
```diff
@@ -80,8 +116,12 @@ public override string ToString()
             var sb = new StringBuilder();
             sb.Append("class KYCVerificationResult {\n");
             sb.Append("  AccountHolder: ").Append(AccountHolder).Append("\n");
-            sb.Append("  BankAccounts: ").Append(BankAccounts).Append("\n");
+            sb.Append("  LegalArrangements: ").Append(LegalArrangements).Append("\n");
```
wboereboom, line 119: These objects are not suitable for printing by default and cause the following errors in LGTM:
`Default 'ToString()': List<KYCLegalArrangementEntityCheckResult> inherits 'ToString()' from 'Object', and so is not suitable for printing.`

Let's see if we do this differently for other object fields. If not; we'll fix it in the model templates later.

wboereboom (changes requested):
Looks good to me! Let's look at the LGTM alerts before we merge.

wboereboom (approved):

AlexandrosMor (approved):

PR #659, opened 2022-10-28
Update HttpURLConnectionClient.cs

BielsasBucket:
**Description**
Add ability to set timeout for RequestAsync, currently defaults to 100000ms. Which is less that Adyen guides suggest for Cloud integrations. 150sec (150000ms) min required.


AlexandrosMor (approved):

BielsasBucket:
Hi,

Just ran the tests with my change removed and they failed just the same. Seems something might be wrong with API key settings as only tests that pass are testing empty, invalid and partially correct API Keys

AlexandrosMor (approved):

Issue #660, opened 2022-10-31
Static ServicePointManager is used to set SecurityProtocol


summerson1985:
**Describe the bug**
Usage of ServicePointManager is not recommended by Microsoft.
ServicePointManager is a globally accessible static class whose properties can be changed by any thread at any time.

When Adyen dot net api library v. 9.1.0 sets ServicePointManager.SecurityProtocol on each request it risks changing behavior for _any other class_ that also relies on ServicePointManager.SecurityProtocol. Think of a dotnet project that integrates with 10 payment service providers each of whom provides their own library. Provider 1 sets ServicePointManager.SecurityProtocol to Tls11, Provider 2 sets it to Tls12 and so on. Under heavy load SecurityProtocol will be changed in an unpredictable and uncontrolled manner allowing supporting teams to have a really fun times debugging the issue.

At our project we use integrations with 4 different providers - one of them is Adyen v.8.2.0 - and I needed to go through the other 3 integrations to make sure they rely on httpClientFactory and httpClient and theoretically should not be affected by Adyen upgrade.

Please consider switching to httpClient: https://learn.microsoft.com/en-us/dotnet/api/system.net.servicepointmanager?view=net-6.0#remarks

**Additional context**
The issue is related to this change: https://github.com/Adyen/adyen-dotnet-api-library/pull/617/files

jillingk:
Hi @summerson1985,

Thank you for reaching out to us with this issue. I see what you are saying, we should realistically find a better solution to set this protocol in a way that does not mess up integrations with other providers. We will create a ticket and look into this!

If you woud like to speed it up and contribute you can also open a PR for this and I can take a look at that immediately. 

Best, Jilling
Adyen

PR #663, opened 2022-11-04
Removed Security protocol TLS1.2

jillingk:
**Description**
#660 Removed setting the security protocol to tls1.2.


wboereboom (approved):

AlexandrosMor:
The recommendation from [microsoft](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca5398) is to not hardcode the `Security protocol`

PR #668, opened 2022-11-11
[PW-7307] Gh pipeline and Makefile to generate Models

jillingk:
**Description**


Adyen/Model/Checkout/Action/IPaymentResponseAction.cs
```diff
@@ -1,4 +1,4 @@
-﻿#region License
+#region License
```
wboereboom, line 1: What actually changed here?

jillingk: Nothing really. I deleted the checkout model directory bc I was testing the generation of the checkout models, but then when merging develop back in this branch it did this for some reason..

Makefile
```diff
@@ -0,0 +1,58 @@
+generator:=csharp
+openapi-generator-cli:=docker run --user $(shell id -u):$(shell id -g) --rm -v ${PWD}:/local -w /local openapitools/openapi-generator-cli:v6.2.1
```
michaelpaul, line 2: Please use the JAR based installation instead of Docker, it's easier to use. See https://github.com/Adyen/adyen-java-api-library/blob/b8ff6ef0f9f49e0f0ed3d6986bb72fb68f527303/Makefile#L3.

michaelpaul (changes requested):

Makefile
```diff
@@ -0,0 +1,70 @@
...
+	mv target/out/src/Adyen.Model/$@/* Adyen/Model/$@
+#adjust this build/src to another file when creating templates
+
+# Checkout spec (and patch version)
+target/spec:
+	git clone https://github.com/Adyen/adyen-openapi.git target/spec
+	perl -i -pe 's/"openapi" : "3.[0-9].[0-9]"/"openapi" : "3.0.0"/' target/spec/json/*.json
+
+# Extract templates (copy them for modifications)
+templates: $(openapi-generator-jar) $(openapi-generator-cli) author template -g $(generator) -o target/templates
```
michaelpaul, line 57:
```suggestion
templates: $(openapi-generator-jar)
	$(openapi-generator-cli) author template -g $(generator) -o target/templates
```

Makefile
```diff
@@ -0,0 +1,70 @@
...
+#6.0.1
+openapi-generator-url:=https://repo1.maven.org/maven2/org/openapitools/openapi-generator-cli/$(openapi-generator-version)/openapi-generator-cli-$(openapi-generator-version).jar
+openapi-generator-jar:=target/openapi-generator-cli.jar
+openapi-generator-cli:=java -jar $(openapi-generator-jar)
+
+
+generator:=csharp
+services:=
+models:=Adyen/Model
+output:=target/out
```
michaelpaul, line 11: Let's make sure this `target` folder is ignored by Git.

michaelpaul (changes requested):

Adyen/Model/Checkout/Action/IPaymentResponseAction.cs
AlexandrosMor: should we include no code changes ? I think we could exclude these files

AlexandrosMor (approved):

michaelpaul (approved):

PR #690, opened 2022-12-14
[PW-7372] Switch HttpWebRequest to HttpClient and Implement RESTfull client

jillingk:
**Description**
Switched the httpclient based on #477, fixed the standard tests and added support for REST which are tested with paymentLinks services. 

**Tested scenarios**
IT


Adyen/HttpClient/HttpClientWrapper.cs
```diff
@@ -0,0 +1,148 @@
...
+//  * ###### ######  #####( ######  #####. ######  #####          #####  ######
+//  * #############  #############  #############  #############  #####  ######
+//  *  ############   ############  #############   ############  #####  ######
+//  *                                      ######
+//  *                               #############
+//  *                               ############
+//  *
+//  * Adyen Dotnet API Library
+//  *
+//  * Copyright (c) 2021 Adyen B.V.
```
AlexandrosMor, line 19:
```suggestion
//  * Copyright (c) 2022 Adyen N.V.
```

Adyen/HttpClient/HttpClientExtension.cs
```diff
@@ -0,0 +1,56 @@
...
+//  * ###### ######  #####( ######  #####. ######  #####          #####  ######
+//  * #############  #############  #############  #############  #####  ######
+//  *  ############   ############  #############   ############  #####  ######
+//  *                                      ######
+//  *                               #############
+//  *                               ############
+//  *
+//  * Adyen Dotnet API Library
+//  *
+//  * Copyright (c) 2021 Adyen B.V.
```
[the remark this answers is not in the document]
wboereboom: Didn't we decide to get rid of the ASCII banners a while back?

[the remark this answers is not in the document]
jillingk: Good point, do I delete all the banners or just the ones of the files I'm changing in this PR? Let's save those trees :)

Adyen.IntegrationTest/CheckoutTest.cs
```diff
@@ -124,12 +125,39 @@ public void PaymentLinksSuccessTest()
...
             Assert.IsNotNull(createPaymentLinkResponse);
             Assert.IsNotNull(createPaymentLinkResponse.Url);
             Assert.IsNotNull(createPaymentLinkResponse.Amount);
             Assert.IsNotNull(createPaymentLinkResponse.Reference);
             Assert.IsNotNull(createPaymentLinkResponse.ExpiresAt);
         }
+        
+        private void PaymentLinksGetSuccessTest(string Id)
+        {   
+            Console.WriteLine();
```
wboereboom, line 139: lingering logging

Adyen/HttpClient/HttpClientExtension.cs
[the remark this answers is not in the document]
wboereboom: Let's do only the ones you changed for now, and we can create separate tickets for the rest to keep this PR readable

Adyen.IntegrationTest/CheckoutTest.cs
jillingk: fixed

wboereboom (dismissed):

wboereboom (dismissed):

Adyen/Client.cs
```diff
@@ -32,41 +33,80 @@ namespace Adyen
 {
     public class Client
     {
-        public Config Config { get; set; }
+        private readonly System.Net.Http.HttpClient _httpClient;
```
AlexandrosMor, line 36: This is something to think if we need to update based on the factory

AlexandrosMor (changes requested):

Adyen/Service/Checkout.cs
```diff
@@ -55,7 +57,7 @@ public Checkout(Client client) : base(client)
             _paymentDetails = new PaymentDetails(this);
             _paymentSession = new PaymentSession(this);
             _paymentsResult = new PaymentsResult(this);
-            _paymentLinksResult = new PaymentLinks(this);
+            _paymentLinksResult = new PaymentLinks(this, null);
```
Kwok-he-Chu, line 60: This might got added due to an upmerge with develop / so the review here may be out-of-scope. @AlexandrosMor @jillingk 

Anyway, I noticed that we're passing a `null` parameter here. So I did some digging as why we would pass `null` here.

Looking at `PaymentLinks.cs`, the parameter is called `endpoint`, which should be named to `linkId` (according to the unittests). Let's rename this for consistency first. Thinking of a use-case, this `linkId` should probably be variable and I recommend not instantiating this when CheckOut is created (unless you want to instantiate a new one every request?)

I think it would be better to pass this linkId elsewhere by restructuring this. This seems like a class where you send a specific request, and then parse its results. Looking at the `getPaymentLinksAsync(...)`, `getPaymentLinks(...)`, `patchPaymentLinks(...)` and `patchPaymentLinksAsync(...)` methods. We will update the request with the right url by appending `/` + `linkId`  

Suggestions:
- [ ] (In a separate ticket*), refactor this `null`-parameter to be outside of its ctor (or don't instantiate it as a memberVariable)
- [ ] Rename the camelCased methods `**G**etPaymentLinksAsync(...)`, `**G**etPaymentLinks(...)`, `**P**atchPaymentLinks(...)` and `**P**atchPaymentLinksAsync(...)` to UpperCased following C# coding guideliness.

Kwok-he-Chu (commented):
Did a review on the changes, great job on the refactor so far, this is not an easy feat!

Another enhancement:
The Github Actions build-and-test give some nice recommendations on several usings to remove, e.g. see: https://github.com/Adyen/adyen-dotnet-api-library/blob/develop/Adyen/Model/Checkout/ServiceError.cs#L25-L42 where we include System twice, which is redundant. We can make it much better :)

Can you create a ticket/issue or just right-click in your IDE and remove/sort unused usings?

Adyen/Service/Checkout.cs
jillingk: Good one about the camelcase (in the other libs we always use camelcase hence my confusion here)

jillingk: The reason I pass that null there is because we use that resource class to create the request url. In case of the Patch and the Get, we need to pass path parameters hence I'm adding that there. It's not a nice way to do it like that, but I just wanted to quickly test the Patch/Get request with the client which is why I did it like that. We'll be sure to refactor this later though when we regenerate the Checkout models and update the services!

AlexandrosMor (dismissed):

Adyen.IntegrationTest/BaseTest.cs
```diff
@@ -151,10 +156,16 @@ protected AdjustAuthorisationRequest CreateAdjustAuthorisationtestRequest(string
...
-            return new Client(apikey, Model.Enum.Environment.Test);
-        }
-        
+            var config = new Config()
+            {
+                XApiKey = ClientConstants.Xapikey,
+                Environment =  Environment.Test
+            };
+            // create basic factory
+            var factory = new ServiceCollection().AddHttpClient().BuildServiceProvider().GetRequiredService<IHttpClientFactory>();
```
Kwok-he-Chu, line 165: I noticed that you're instantiating a new service collection and getting a real httpclientfactory here, I recommend just passing a new HttpClient() here to get a similar effect so you do not need to include `using Microsoft.Extensions.DependencyInjection` the DI dependencyInjection extensions packages.

What do you think?

Adyen.IntegrationTest/CheckoutTest.cs
```diff
@@ -69,10 +72,10 @@ public void PaymentsTest()
         {
             var paymentResponse = _checkout.Payments(CreatePaymentRequestCheckout());
             Assert.IsNotNull(paymentResponse.AdditionalData);
-            Assert.AreEqual("411111", paymentResponse.AdditionalData["cardBin"]);
```
Kwok-he-Chu, line 72: Noticed some rename-changes here, quick question, could you confirm this?

- [ ] cardBin is removed, renamed to cardSummary I assume?

- [ ] paymentMethodVariant is removed, renamed to paymentMethod I assume?

Adyen.Test/HeaderRequestTest.cs
```diff
@@ -30,90 +33,80 @@ namespace Adyen.Test
...
-        {
-            _httpUrlConnectionClient = new HttpUrlConnectionClient();
-        }
+        
         [TestMethod]
         public void BasicAuthenticationHeadersTest()
         {
-            var httpWebRequest = _httpUrlConnectionClient.GetHttpWebRequest(_endpoint, MockPaymentData.CreateConfingMock(), false);
-
-            Assert.IsNotNull(httpWebRequest.UserAgent);
```
Kwok-he-Chu, line 46: Doing a quick before/after check: I see 2 (potentially 3) mismatches in left-out checks. Just checking if it was done intentionally.

- [x] Assert.IsNotNull(httpWebRequest.UserAgent);

- [x] Assert.AreEqual(httpWebRequest.Address.ToString(), _endpoint);
 Assert.AreEqual<string>(httpRequestMessage.RequestUri?.ToString(), _endpoint); // I assume this is the same :)

- [x] Assert.AreEqual(httpWebRequest.Headers["Accept-Charset"], "UTF-8");

- [x] Assert.AreEqual(httpWebRequest.Headers["Cache-Control"], "no-cache");

- [ ] Assert.AreEqual(httpWebRequest.ContentType, "application/json");
Missing (?)

- [x] Assert.IsNotNull(httpWebRequest.Headers["Authorization"]);

- [ ] Assert.IsTrue(httpWebRequest.UseDefaultCredentials);
Missing

- [x] Assert.AreEqual(httpWebRequest.Headers["Cache-Control"], "no-cache");

- [ ] Assert.IsNull(httpWebRequest.Headers["x-api-key"]);
Missing (?)

Kwok-he-Chu (commented):
I went over some points that stood out to me and several enhancements that could be applied.
Left some questions as well for clarification, thanks :) 👍

Adyen/Config.cs
```diff
@@ -34,6 +34,7 @@ public class Config
         public string Password { get; set; }
```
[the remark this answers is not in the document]
jillingk: @Kwok-he-Chu, we already had some discussion about this actually and I think our conclusion was that we do want to keep it for now even though it's really not our preferred way!

Adyen.Test/RecurringTest.cs
```diff
@@ -95,13 +94,13 @@ public void TestDisable803()
                 var disableResult = recurring.Disable(disableRequest);
                 Assert.Fail("Exception expected!");
             }
-            catch (HttpClientException exception)
+            catch (Exception exception)
             {
-                Assert.AreNotEqual(200, exception.Code);
+                Assert.AreNotEqual<int>(200, exception.GetHashCode());
```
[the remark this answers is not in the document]
jillingk: Good catch. I commented out these tests because the exceptions of the new client were different, but since we're wrapping the new exceptions in our old Adyen exception they can be added again :)

Adyen.IntegrationTest/BaseTest.cs
jillingk: Fair point, it was just to test the overloaded constructor with the factory really. Since we know it works we might as well remove the DI extension though.

Adyen.IntegrationTest/CheckoutTest.cs
jillingk: Not too sure actually, as I'm seeing both in the specs under AdditionalCardData. Ditto for the paymentMethod, both still in the specs as well. But seeing as I don't encounter them anymore whilst debugging and getting the raw responses on the checkout IT's, I suspect they retired them (partly).

Adyen.Test/HeaderRequestTest.cs
jillingk: Fixed it

Adyen.Test/HeaderRequestTest.cs
```diff
@@ -30,90 +33,81 @@ namespace Adyen.Test
...
+            var client = new HttpClientWrapper(MockPaymentData.CreateConfingMock(), new System.Net.Http.HttpClient());
+            var httpRequestMessage = client.GetHttpRequestMessage(_endpoint, false, "requestBody", null, HttpMethod.Post);
+            
+            Assert.IsNotNull(httpRequestMessage.Headers.UserAgent);
+            Assert.AreEqual(httpRequestMessage.RequestUri?.ToString(), _endpoint);
+            Assert.AreEqual(httpRequestMessage.Headers.AcceptCharset.ToString(), "UTF-8");
+            Assert.AreEqual(httpRequestMessage.Headers.CacheControl?.ToString(), "no-cache");
+            Assert.AreEqual(httpRequestMessage.Content.Headers.GetValues("Content-Type").FirstOrDefault(), "application/json; charset=utf-8");
+            Assert.IsNotNull(httpRequestMessage.Headers.Authorization);
+            httpRequestMessage.Headers.GetValues("x-api-key");
```
Kwok-he-Chu, line 51: Recommendation, use: Assert.IsFalse(httpRequestMessage.Headers.TryGetValues("x-api-key", out var _));

Kwok-he-Chu: Remove: [ExpectedException(typeof(InvalidOperationException), "The given header was not found.")]

Adyen/HttpClient/Interfaces/IClient.cs
```diff
@@ -15,24 +15,43 @@
...
-        string Post(string endpoint, Dictionary<string, string> postParameters, Config config);
+        /// <summary>
+        /// Request Method asynchronous.
+        /// </summary>
+        /// <param name="endpoint">Url of the endpoint.</param>
+        /// <param name="json">Json request parameters for Post/Patch.</param>
+        /// <param name="isApiKeyRequired"></param>
+        /// <param name="requestOptions">Optional parameter used to specify the options for the request.</param>
+        /// <param name="httpMethod">Request Method.</param>
+        /// <returns>string</returns>
```
Kwok-he-Chu, line 53: Suggestion `<returns>A task with string response in json format</returns>`

Adyen/HttpClient/Interfaces/IClient.cs
```diff
@@ -15,24 +15,43 @@
...
+        string Request(string endpoint, string json, bool isApiKeyRequired, RequestOptions requestOptions = null, HttpMethod httpMethod = null);
         
-        Task<string> RequestAsync(string endpoint, string json, Config config, bool isApiKeyRequired, RequestOptions requestOptions);
-        string Post(string endpoint, Dictionary<string, string> postParameters, Config config);
+        /// <summary>
+        /// Request Method asynchronous.
+        /// </summary>
+        /// <param name="endpoint">Url of the endpoint.</param>
+        /// <param name="json">Json request parameters for Post/Patch.</param>
+        /// <param name="isApiKeyRequired"></param>
```
Kwok-he-Chu, line 50: Add description for `isApiKeyRequired`

Adyen/HttpClient/Interfaces/IClient.cs
```diff
@@ -15,24 +15,43 @@
...
 
 namespace Adyen.HttpClient.Interfaces
 {
-    public interface IClient
+    public interface IClient : IDisposable
     {
-        string Request(string endpoint, string json, Config config);
-        string Request(string endpoint, string json, Config config, bool isApiKeyRequired, RequestOptions requestOptions);
+        /// <summary>
+        /// Request Method synchronous.
```
Kwok-he-Chu, line 35: Suggestion for summary: `Sends synchronous request to the specified <param refname="endpoint"/>.`

Adyen/HttpClient/Interfaces/IClient.cs
```diff
@@ -15,24 +15,43 @@
...
+        /// <param name="isApiKeyRequired"></param>
+        /// <param name="requestOptions">Optional parameter used to specify the options for the request.</param>
+        /// <param name="httpMethod">Request Method.</param>
+        /// <returns>string</returns>
+        string Request(string endpoint, string json, bool isApiKeyRequired, RequestOptions requestOptions = null, HttpMethod httpMethod = null);
         
-        Task<string> RequestAsync(string endpoint, string json, Config config, bool isApiKeyRequired, RequestOptions requestOptions);
-        string Post(string endpoint, Dictionary<string, string> postParameters, Config config);
+        /// <summary>
+        /// Request Method asynchronous.
```
Kwok-he-Chu, line 46: Suggestion for summary: `Sends asynchronous request to the specified <param refname="endpoint"/>.`

Kwok-he-Chu (commented):
Review

Adyen.Test/HeaderRequestTest.cs
```diff
@@ -30,90 +33,81 @@ namespace Adyen.Test
...
         private readonly string _endpoint = "https://endpoint:8080/";
-
-        public HeaderRequestTest()
-        {
-            _httpUrlConnectionClient = new HttpUrlConnectionClient();
-        }
+        
         [TestMethod]
+        [ExpectedException(typeof(InvalidOperationException), "The given header was not found.")]
         public void BasicAuthenticationHeadersTest()
```
Kwok-he-Chu, line 40: This is a **very** nice to have, you can open a different ticket for this, if you'd like:

[Enhancement] Rename tests to consistently identify what it-is that you're testing following the "Given _ [MethodName] _ When_ [Condition]_ ResultShouldBe _ [ExpectedResult]" convention.
See: https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#naming-your-tests

This way, it will be clear what is being tested (and you can get rid of unnecessary checks / refactor the unittests into smaller bits, e.g. > do we really need to test the utf-8?)


Example from this PR: `BasicAuthenticationHeadersTest` is testing whether the AuthenticationHeaders are included or not, the same applies to `ApiKeyBasedHeadersTest` where we include "x-api-key" in our headers or not.

Looking at the business logic for this, I expect four tests:
```csharp
"AuthenticationHeaders":     provided | "x-api-key": provided     // Resulting headers["x-api-key"] should be not null
"AuthenticationHeaders":     provided | "x-api-key": not-provided // Result Authentication headers should be not null
"AuthenticationHeaders": not-provided | "x-api-key": provided     // Resulting headers["x-api-key"] should be not null
"AuthenticationHeaders": not-provided | "x-api-key": not-provided // Resulting headers["x-api-key"] should be null, result authentication headers should be null or  even throw an error here as no authentication is provided.
```

Example: `Given_GetHttpWebRequest_When_AuthenticationHeaders_Is_Provided_ResultShouldBe_XXXX`

This covers all cases for the business logic in these lines: https://github.com/Adyen/adyen-dotnet-api-library/pull/690/files#diff-8b602ca15935ec373df692637c85201512137f432ab908df44fa25ad74e8467bR107-R117
```csharp
  if (isApiKeyRequired || _config.HasApiKey)
  {
      httpWebRequest.Headers.Add("x-api-key", _config.XApiKey);
  }
  else if (_config.HasPassword)
  {
      // ...
  }
```

Adyen/HttpClient/Interfaces/IClient.cs
Kwok-he-Chu: Related PR #699, you can ignore this I think ^

wboereboom (dismissed):

Kwok-he-Chu (approved):
Great job.

wboereboom (approved):

michaelpaul (approved):

Kwok-he-Chu (approved):

PR #703, opened 2023-01-03
[PW-7308] Generate (Classic) Payments, Add Missing Endpoints And Bump Version

jillingk:
**Description**
This PR bumps the classic payments to v68. It also generated new models for payments and changed services and tests to utilise them. Furthermore, it adds the missing endpoints for the classic modifications API (also pal). 

**Tested scenarios**
IT and UT's


wboereboom:
Wondering whether we should create a new service with all the classic endpoints rather than keeping modification and payment separate. 

If we ever automate service generation these endpoints will end up in the same service as they are part of the same spec file.

Adyen/Model/Checkout/CreateCheckoutSessionRequest.cs
```diff
@@ -249,12 +249,28 @@ public CreateCheckoutSessionRequest(AccountInfo accountInfo = default(AccountInf
             this.ShopperReference = shopperReference;
             this.ShopperStatement = shopperStatement;
             this.SocialSecurityNumber = socialSecurityNumber;
-            this.SplitCardFundingSources = splitCardFundingSources;
+            // use default value if no "splitCardFundingSources" provided
+            if (splitCardFundingSources == null)
```
michaelpaul, line 253: This change should be in the Checkout PR. Or not even there if it didn't came from the generator.

AlexandrosMor (approved):

michaelpaul (approved):

PR #705, opened 2023-01-03
Fix Async Tests

Kwok-he-Chu:
**Description**

- CheckoutTest now throws and asserts the InvalidOperationException properly, instead of catching the exception using Attributes

- MarketPayAccountTest now calls the async method properly

- Assigning bool values in CreateCheckoutSessionRequest can never be null, just assign the variables directly

- InvalidDataException can now be thrown, by comparing it to the defaultEnum in GenericIssuerPaymentMethodDetails.cs instead of null

- FrequencyEnum can throw InvalidDataException by comparing it to the default(FrequencyEnum) as it could never be null in Mandate.cs

- TestSuspendAccountHolderSuccessAsync now calls its async request in MarketpayAccountTest.cs

### DIFF
--- a/Adyen.Test/CheckoutTest.cs
+++ b/Adyen.Test/CheckoutTest.cs
@@ -23,19 +23,19 @@
 
 #endregion
 
+using Adyen.Model;
 using Adyen.Model.ApplicationInformation;
 using Adyen.Model.Checkout;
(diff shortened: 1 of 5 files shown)


wboereboom (approved):
Thanks for contributing :D

PR #708, opened 2023-01-05
[PW-7313] Generate Checkout Models and update services/tests

jillingk:
**Description**
Generated Checkout Models with csharp-netcore generator (also slightly modified them). Updated services in the checkout class and added ApplePaySessions. Lastly it fixed all the UT/IT's.

**Tested scenarios**
UT/IT


Makefile
```diff
@@ -4,8 +4,8 @@ openapi-generator-jar:=target/openapi-generator-cli.jar
 openapi-generator-cli:=java -jar $(openapi-generator-jar)
 
 
-generator:=csharp
-services:=LegalEntityManagement
```
wboereboom, line 8: Why remove LEM?

Makefile
```diff
@@ -4,8 +4,8 @@ openapi-generator-jar:=target/openapi-generator-cli.jar
...
+generator:=csharp-netcore
+services:=Checkout
```
jillingk, line 8:
```suggestion
services:=Checkout LegalEntityManagement
```

wboereboom (approved):

AlexandrosMor (approved):

PR #713, opened 2023-01-09
PW-7317: Management API

michaelpaul:
**Description**

Add support for [Management API](https://docs.adyen.com/api-explorer/Management/1/overview).

**TODO**

- [x] Friendly method names
- [x] Add to README.md after #714


Makefile
```diff
@@ -20,6 +20,7 @@ Payments: spec=PaymentService-v68
 Recurring: spec=RecurringService-v68
 Payouts: spec=PayoutService-v68
 Management: spec=ManagementService-v1
+Management: service=management
```
wboereboom, line 23: Are we going to have two entries for every model and service per API. Is there another way to set this up?

wboereboom (dismissed):

Makefile
michaelpaul: This is only necessary for services where we generate both models and the client.

wboereboom: But that will probably be all of them in due time

jillingk (approved):

AlexandrosMor (approved):

PR #724, opened 2023-01-23
[PW-7542] Nullable integers to fix default value of int (0) being emitted

jillingk:
I encountered the same problem we have had before of integers being emitted during serialization if you set it to zero. This is caused by the fact that the default value of integers in csharp (and most languages) is equal to zero. This is easy fixable by making all integers nullable (which I did on spec level). Next to that I also found that we serialize a bunch of parameters to its default value if not filled in, which I fixed in this PR by adjusting the specs as well.


Adyen.Test/PaymentTest.cs
```diff
@@ -173,6 +173,16 @@ public void TestPaymentRequest3DApplicationInfo()
...
             Assert.AreEqual(paymentRequest.ApplicationInfo.AdyenLibrary.Version, ClientConfig.LibVersion);
         }
+        
+        [TestMethod]
+        public void TestCaptureDelaySerialization()
+        {
+            var paymentRequest = MockPaymentData.CreateFullPaymentRequest3D();
+            string jsonString = paymentRequest.ToJson();
+            Console.WriteLine(jsonString);
+            Assert.IsTrue(jsonString.Contains("\"captureDelayHours\": 0,"));
```
wboereboom, line 183: 🚀 👍

templates/csharp/modelGeneric.mustache
```diff
@@ -199,15 +199,15 @@
...
         {{#isDate}}
         [JsonConverter(typeof(OpenAPIDateConverter))]
         {{/isDate}}
         {{#deprecated}}
         [Obsolete]
         {{/deprecated}}
-        public {{{dataType}}} {{name}} { get; {{#isReadOnly}}private {{/isReadOnly}}set; }
-
+        public {{{dataType}}}{{#isInteger}}?{{/isInteger}} {{name}} { get; {{#isReadOnly}}private {{/isReadOnly}}set; }
+
```
wboereboom, line 210: Can we remove this tab and regenerate?

Adyen/Model/Payments/ThreeDSecureData.cs
```diff
@@ -64,11 +66,12 @@ public enum AuthenticationResponseEnum
 
         }
 
+
```
wboereboom, line 69: can we get rid of this additional blank line?

wboereboom (commented):
two minor comments to do with formatting. Looks great otherwise. Like the added unit tests as well 👍 
Great work

templates/csharp/modelGeneric.mustache
jillingk: fixed

Adyen/Model/Payments/ThreeDSecureData.cs
jillingk: and fixed

AlexandrosMor (approved):

Adyen.Test/CheckoutTest.cs
```diff
@@ -674,14 +674,14 @@ public void DragonpayDetailsDeserializationTest()
             Assert.AreEqual(result.Type, DragonpayDetails.TypeEnum.Ebanking);
         }
 
-        [TestMethod]
+        /*[TestMethod]
```
wboereboom, line 677: commented code

Makefile
```diff
@@ -36,6 +36,7 @@ $(services): target/spec $(openapi-generator-jar)
 		-t templates/csharp \
 		-o $(output) \
 		--model-package $@ \
+		--skip-validate-spec \
```
wboereboom, line 39: why do we add this flag?

jillingk: The generator performs a validation on the specs it tries to build on. Since our specs are a bit wonky, it would throw errors and not complete generation. Hence the this flag :)

Adyen.Test/CheckoutTest.cs
jillingk: This payment method was removed in checkout, I'll delete the test instead of the comment

michaelpaul (approved):

AlexandrosMor (approved):

PR #726, opened 2023-01-25
Add missing values to ResultCodeEnum

amitin-boden:
**Description**
Currently it's not always possible to use models defined in the lib for JSON parsing, because the ResultCodeEnum enum contains not all the possible values. This change should fix the issue.

### DIFF
--- a/Adyen/Model/Enum/ResultCodeEnum.cs
+++ b/Adyen/Model/Enum/ResultCodeEnum.cs
@@ -39,7 +39,8 @@ namespace Adyen.Model.Enum
     {
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


jillingk:
Hi @amitin-boden,

Thanks a lot for your contribution! However we are actually in the process of deprecating this shared enum folder and generating all the endpoint specific (enum)models for most of our API's. If you look through the models in our develop branch you should be able to find an extensive list of enum classes, fully up to date, which we will try to release very soon. Please do let me know though if you find any models not yet up to date, as we want to make this library as complete as possible.

Best, 
Jilling
Adyen

PR #727, opened 2023-01-27
Make services implement their interfaces

karol-gro:
**Description**

These two services:
- PosPaymentCloudApi
- PosPaymentLocalApi

have their matching interfaces, but don't implement them. This was introduced in #578, where interfaces were added to source code, but not implemented

**Tested scenarios**

No changes done, which may influence behavior

### DIFF
--- a/Adyen/Service/IPosPaymentLocalApi.cs
+++ b/Adyen/Service/IPosPaymentLocalApi.cs
@@ -22,6 +22,7 @@
 #endregion
 
 using System.Net.Security;
+using System.Threading.Tasks;
 using Adyen.Model.Nexo;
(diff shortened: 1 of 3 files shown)


karol-gro:
I noticed the integration tests failed on CI. The reason is that the build uses secrets, which are unavailable for security reasons on PRs from forks

wboereboom:
Hi @karol-gro,

Thanks for contributing.
This implementation must've slipped through somehow, so I'm glad you found and fixed it!

Kind Regards,
Wouter
Adyen

wboereboom (approved):

AlexandrosMor (approved):

PR #728, opened 2023-01-27
[PW-7958] Generic Type in Service to reduce boiler plate code

jillingk:
**Description**
Adds support for generic response type in the ServiceResource class to reduce boilerplate code.
**Tested scenarios**
UT and IT

### DIFF
--- a/Adyen/Service/Checkout.cs
+++ b/Adyen/Service/Checkout.cs
@@ -25,7 +25,6 @@ using System.Net.Http;
 using System.Threading.Tasks;
 using Adyen.Model.Checkout;
 using Adyen.Service.Resource.Checkout;
-using Newtonsoft.Json;
 using CardDetails = Adyen.Service.Resource.Checkout.CardDetails;
 using PaymentDetails = Adyen.Service.Resource.Checkout.PaymentDetails;
 using PaymentRequest = Adyen.Model.Checkout.PaymentRequest;
@@ -76,8 +75,7 @@ namespace Adyen.Service
         public PaymentResponse Payments(PaymentRequest paymentRequest, RequestOptions requestOptions = null)
         {
             var jsonRequest = paymentRequest.ToJson();
-            var jsonResponse = _payments.Request(jsonRequest, requestOptions);
-            return JsonConvert.DeserializeObject<PaymentResponse>(jsonResponse);
+            return _payments.Request<PaymentResponse>(jsonRequest, requestOptions);
         }
 
         /// <summary>
@@ -89,8 +87,7 @@ namespace Adyen.Service
         public async Task<PaymentResponse> PaymentsAsync(PaymentRequest paymentRequest, RequestOptions requestOptions = null)
         {
             var jsonRequest = paymentRequest.ToJson();
-            var jsonResponse = await _payments.RequestAsync(jsonRequest, requestOptions);
-            return JsonConvert.DeserializeObject<PaymentResponse>(jsonResponse);
+            return await _payments.RequestAsync<PaymentResponse>(jsonRequest, requestOptions);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Adyen/Service/Checkout.cs
```diff
@@ -293,8 +292,7 @@ public CreateCheckoutSessionResponse Sessions(CreateCheckoutSessionRequest creat
         public async Task<CreateCheckoutSessionResponse> SessionsAsync(CreateCheckoutSessionRequest createCheckoutSessionRequest)
         {
             var jsonRequest = createCheckoutSessionRequest.ToJson();
-            var jsonResponse = await _sessions.RequestAsync(jsonRequest);
```
wboereboom, line 296: Why not update the entire class? That way we could also maybe remove the dependency for this from this class

AlexandrosMor: Indeed it is only in one place :(

jillingk: Ah yeah I just wanted to test it and thought we'd change the rest later during the service automation. I can change it manually as well though!

jillingk: Changed it :)

Kwok-he-Chu (commented):
I've given my input on the PR. Good work on it so far. I think we can do even better. One thing that came up to my mind was: Do we need a separate `ServiceResource`? The logic seems to be part of a client implementation that does the actual request. Something like this:

```csharp
public interface ISomeClient.cs // The implemementation should have the HttpClient that performs the actual request.
{
     Task<T> RequestAsync(string json, RequestOptions requestOptions = null, HttpMethod httpMethod = null, CancellationToken cancellationToken = default);
     T Request(string json, RequestOptions requestOptions = null, HttpMethod httpMethod = null, CancellationToken cancellationToken = default);
}
```


**Edit:** I just noticed that ServiceResource acts like an abstract class for all the services (for example [this](https://docs.google.com/document/d/1wGmZRS1SXGQgLsDkmQl_OEJYHhqCNS99stdxFZWoMco/edit). We should consider rename the class (by adding a suffix) to reflect its behavior. If you decide on refactoring this part, we should pick it up in a different ticket. I think all is good for now 👍

Adyen/Service/ServiceResource.cs
```diff
@@ -4,6 +4,7 @@
 using System.Net.Security;
 using System.Threading.Tasks;
 using Adyen.Model;
+using Newtonsoft.Json;
 
 namespace Adyen.Service
```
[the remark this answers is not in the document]
Kwok-he-Chu: I noticed that we're introducing newtonsoft.json here, can we consider using System.Text.Json?

_Note: I'm aware that the library already uses Newtonsoft.jJon, if that's the only thing that you're using right now, I would keep it consistent, keep using Newtonsoft.Json and refactor this when the team is ready to move to Systen.Text.Json._

Adyen/Service/ServiceResource.cs
```diff
@@ -23,11 +24,25 @@ public string Request(string json, RequestOptions requestOptions = null, HttpMet
...
+            return JsonConvert.DeserializeObject<T>(jsonResponse);
+        }
 
         public Task<string> RequestAsync(string json, RequestOptions requestOptions = null, HttpMethod httpMethod = null)
         {
             var clientInterface = _abstractService.Client.HttpClient;
             return clientInterface.RequestAsync(Endpoint, json, requestOptions, httpMethod);
         }
+        
+        public async Task<T> RequestAsync<T>(string json, RequestOptions requestOptions = null, HttpMethod httpMethod = null)
```
Kwok-he-Chu, line 41: Pass on a CancellationToken here for asynchronous method.

```csharp
 public async Task<T> RequestAsync<T>(string json, RequestOptions requestOptions = null, HttpMethod httpMethod = null, CancellationToken cancellationToken = default)
```

michaelpaul (approved):

AlexandrosMor (approved):

PR #730, opened 2023-02-01
[PW-7805] Timeout variable for HttpClient

jillingk:
**Description**
This PR adds a timeout variable in the config, that sets the timeout for the client constructors with a default httpclient instance. 

**Tested scenarios**
Since the HttpClient object is private readonly in the HttpClientWrapper object, it's hard to test with UT. However I verified with a debugger that the timeout is correctly set.

### DIFF
--- a/Adyen/Client.cs
+++ b/Adyen/Client.cs
@@ -57,7 +57,10 @@ namespace Adyen
             
             // Set Timeout for HttpClient
             var httpClient = new System.Net.Http.HttpClient();
-            httpClient.Timeout = Config.Timeout == default ? TimeSpan.FromMilliseconds(100000) : TimeSpan.FromMilliseconds(Config.Timeout);
+            if (Config.Timeout != default)
+            {
+                httpClient.Timeout = TimeSpan.FromMilliseconds(Config.Timeout);
+            }
             
             _client = new HttpClientWrapper(Config, httpClient);
         }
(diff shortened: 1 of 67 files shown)


Adyen/Client.cs
```diff
@@ -53,7 +54,12 @@ public Client(string username, string password, Environment environment, string
                 Environment = environment
             };
             SetEnvironment(environment, liveEndpointUrlPrefix);
-            _client = new HttpClientWrapper(Config, new System.Net.Http.HttpClient());
+            
+            // Set Timeout for HttpClient
+            var httpClient = new System.Net.Http.HttpClient();
+            httpClient.Timeout = Config.Timeout == default ? TimeSpan.FromMilliseconds(100000) : TimeSpan.FromMilliseconds(Config.Timeout);
```
michaelpaul, line 60: Let's assign a timeout only if the user explicitly provides one. 
```suggestion
            if (Config.Timeout != default)
            {
                httpClient.Timeout = TimeSpan.FromMilliseconds(Config.Timeout);
            }

```

michaelpaul (changes requested):

Adyen/Client.cs
jillingk: Really wanted to use a one liner here and the default is 100000 milliseconds haha

Adyen/Client.cs
```diff
@@ -66,14 +75,30 @@ public Client(string xapikey, Environment environment, string liveEndpointUrlPre
...
+            _client = new HttpClientWrapper(Config, httpClient);
         }
 
         public Client(Config config)
         {
             Config = config;
             SetEnvironment(Config.Environment, Config.LiveEndpointUrlPrefix);
-            _client = new HttpClientWrapper(Config, new System.Net.Http.HttpClient());
+            
+            // Set Timeout for HttpClient
```
wboereboom, line 94: can we move this code to a helper function rather than copying it three times?

jillingk: Done

michaelpaul (approved):

AlexandrosMor (approved):

PR #731, opened 2023-02-02
[PW-7996] StoredValue API endpoints

jillingk:
**Description**
This PR adds support for the StoredValue endpoints. The structure is a little bit different than how we usually add them (that is now without the resource class), so let me know how this looks like. I also opted for making Async methods have the normal method name and adjust the Sync ones. 

**Tested scenarios**
UT's and One IT to test the correctness of the post url.


jillingk:
This PR is also chained on [this one](https://github.com/Adyen/adyen-dotnet-api-library/pull/728), to use the generic types so we make the service class more readable. So please approve that one first :)

Adyen/Service/StoredValue.cs
```diff
@@ -0,0 +1,155 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Adyen.Model.StoredValue;
+
+namespace Adyen.Service
+{
+    public class StoredValue : AbstractService
+    {
+        private readonly string _baseUrl;
```
wboereboom, line 10:
```suggestion
        private static final string baseUrl = client.Config.StoredValueEndpoint + "/" + Constants.ClientConfig.StoredValueVersion;;
```

jillingk: This doesn't work actually, as the client.config is passed from the constructor and thus cannot be initialised beforehand..

michaelpaul (dismissed):

AlexandrosMor (approved):

AlexandrosMor (dismissed):

AlexandrosMor (approved):

wboereboom (approved):

PR #739, opened 2023-02-17
Update schema and add System.Text.Json support for Notification Requests

JakubK:
**Description**

The purpose of this PR is to make it accessible for ASP.NET WebAPI's using `System.Text.Json`.
After generating blank ASP.NET WebAPI with such a dummy Webhook Controller Action:

```csharp
[HttpPost]
public async Task<ActionResult> AdyenConfirmAsync([FromBody] NotificationRequest request)
{
    return Ok(request);
}
```

I noticed that most of the fields in payload are missing. Short session with debugger helped me investigate that there were the following issues:

1. `NotificationRequestItem.Success` field was of type boolean while the Adyen sends it as string.
2. `NotificationRequest.NotificationItemContainers` field was improperly mapped because in Json payload it's named as `notificationItems`.
3.  `NotificationRequestItemContainer.NotificationItem` field was improperly mapped because in Json payload it's named `NotificationRequestItem`.

To solve issue `1` I have changed the type to `string`.
I have added `System.Text.Json` and used it's attribute equivalent to `JsonProperty` of `Newtonsoft.Json` in order to fix last 2 isues.

My previous commit has more invasive solution for this problem by renaming fields, but I've decided that this will be better.

**Tested scenarios**
After making these changes my test WebAPI could get the payload and I could browse it in debugger. Unit tests were also updated and they are green, I was not able to run integration tests neither on my branch or on develop.

**Fixed issue**:  
I did not find any related issue on this repository

### DIFF
--- a/Adyen.Test/NotificationTest.cs
+++ b/Adyen.Test/NotificationTest.cs
@@ -41,10 +41,10 @@ namespace Adyen.Test
             var jsonRequest = MockFileToString(mockPath);
             var notificationHandler = new NotificationHandler();
             var handleNotificationRequest = notificationHandler.HandleNotificationRequest(jsonRequest);
-            var notificationRequestItemContainer = handleNotificationRequest.NotificationItems.FirstOrDefault();
+            var notificationRequestItemContainer = handleNotificationRequest.NotificationItemContainers.FirstOrDefault();
             if (notificationRequestItemContainer == null)
                 Assert.Fail("NotificationRequestItemContainer is null");
-            var notificationItem = notificationRequestItemContainer.NotificationRequestItem;
+            var notificationItem = notificationRequestItemContainer.NotificationItem;
             Assert.AreEqual("AUTHORISATION", notificationItem.EventCode);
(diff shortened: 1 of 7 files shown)


jillingk:
Hi @JakubK,

Thanks a lot for reaching out to us and wanting to contribute. Also massive apologies for the radio silence on our part. 

Quick question, wouldn't it be possible to just install the AspNetCore Newntonsoft.Json package and use it in the webhook controller as such? I'm suspecting that this should solve your issue as we're explicitly depending on Newtonsoft for serialization in this library and not the default System.Text.Json. 

On a side note, we are also investigating if we can migrate to System.Text.Json serialization. However if we decide to do so in the future, we will probably want to do it at once and not introduce both packages in the library.

Best Jilling
Adyen

JakubK:
Hi, 
Yeah I think this usage of Newtonsoft just for the webhook could be possible.
For now I have recreated the mentioned model classes without the `Newtonsoft.Json` decorators and with proper data types.
I guess I'll have to just pick this or the idea you suggested. Thanks for the reply and I'm looking forward for migration to System.Text.Json!

PR #740, opened 2023-02-20
[ITT-366] Regenerated Payout models/services

jillingk:
**Description**
This PR regenerates the Payout models and services. 

**Tested scenarios**
UT/IT


wboereboom (commented):
Looks good. Blocked by method naming for now.

wboereboom (dismissed):

Adyen.Test/BaseTest.cs
```diff
@@ -431,36 +431,6 @@ protected Client CreateMockTestClientPosLocalApiRequest(string fileName)
             return clientMock;
         }
 
-        /// <summary>
-        /// Creates mock test client errors
-        /// </summary>
-        /// <param name="status"></param>
-        /// <param name="fileName"></param>
-        /// <returns></returns>
-        protected Client CreateAsyncMockTestClientForErrors(int status, string fileName)
```
Kwok-he-Chu, line 440: Just checking whether this removal is intended

Adyen/Service/Payout/ReviewingService.cs
```diff
@@ -0,0 +1,86 @@
...
+    /// <summary>
+    /// Represents a collection of functions to interact with the API endpoints
+    /// </summary>
+    public class ReviewingService : AbstractService
+    {
+        private readonly string _baseUrl;
+        
+        public ReviewingService(Client client) : base(client)
+        {
+            _baseUrl = client.Config.Endpoint + "/pal/servlet/Payout/" + ClientConfig.PayoutApiVersion;
```
Kwok-he-Chu, line 34: Double-checking here, how do we distinguish between the live and test endpoints?

Adyen.Test/PayoutTest.cs
```diff
@@ -33,11 +36,11 @@ public class PayoutTest : BaseTest
         [TestMethod]
         public void StoreDetailAndSubmitThirdPartySuccessTest()
         {
-            var client = CreateMockTestClientNullRequiredFieldsRequest("Mocks/payout/storeDetailAndSubmitThirdParty-success.json");
-            var payout = new Payout(client);
+            var client = CreateMockTestClientApiKeyBasedRequestAsync("Mocks/payout/storeDetailAndSubmitThirdParty-success.json");
```
Kwok-he-Chu, line 39: We're using the client creation method here for `CreateMockTestClientApiKeyBasedRequestAsync`, but the payout.StoreDEtailsAndSubmitPayout is calling the synchronous method, I assume that this test passes because we indirectly call the asynchronous call (since it indirectly calls the async version). Could you verify whether the test succeeds/fails? 

The main difference is in its setup, see snippet in BaseTest.cs:
```csharp
/// diff. between CreateMockTestClientNullRequiredFieldsRequest and CreateMockTestClientApiKeyBasedRequestAsync

/// CreateMockTestClientApiKeyBasedRequestAsync
   var confMock = MockPaymentData.CreateConfigApiKeyBasedMock(); // diff 1
    ClientInterfaceMock.Setup(x => x.RequestAsync(It.IsAny<string>(),
        It.IsAny<string>(), It.IsAny<RequestOptions>(), It.IsAny<HttpMethod>())).ReturnsAsync(response); 
   var confMock = MockPaymentData.CreateConfigApiKeyBasedMock(); // diff 2
              
/// CreateMockTestClientNullRequiredFieldsRequest 
      clientInterfaceMock // diff 1
          .Setup(x => x.Request(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<RequestOptions>(), null))
          .Returns(response);      
      clientInterfaceMock // diff 1
          .Setup(x => x.RequestAsync(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<RequestOptions>(), null))
          .Returns(Task.FromResult(response));              
      var confMock = MockPaymentData.CreateConfigMock(); // diff 2
```

Kwok-he-Chu (commented):
Gj! Added some comments since we regenerated a bunch of models, I'd like to check whether some of the changes are intentional.

Adyen/Model/Payout/Amount.cs
```diff
@@ -24,6 +23,7 @@
 using Newtonsoft.Json.Converters;
 using Newtonsoft.Json.Linq;
 using System.ComponentModel.DataAnnotations;
+using OpenAPIDateConverter = Adyen.ApiSerialization.OpenAPIDateConverter;
```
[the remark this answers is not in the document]
jillingk: Yep, it's for DateTime objects like date of birth, where you don't need the full iso string as dd/mm/yyyy is enough.

Adyen/Service/Payout/ReviewingService.cs
jillingk: client.config.Endpoint is an config object set when constructing the Client object and either set to the live or test endpoint of the specific endpoint url.

Adyen.Test/PayoutTest.cs
jillingk: Yes indeed. Since every method now is pretty much an async method being awaited I have to use async mock tests for them as well. They all pass after converting them to this mock client.

Adyen/Model/Payout/Address.cs
```diff
@@ -61,28 +61,28 @@ public Address(string city = default(string), string country = default(string),
         /// The name of the city. Maximum length: 3000 characters.
         /// </summary>
         /// <value>The name of the city. Maximum length: 3000 characters.</value>
-        [DataMember(Name = "city", IsRequired = false, EmitDefaultValue = true)]
+        [DataMember(Name = "city", IsRequired = false, EmitDefaultValue = false)]
```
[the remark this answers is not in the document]
jillingk: We mostly rely on the API to tell us what we need to add to our requests and what's missing. So I think that is why we should go for omitting all default values. As far as the empty address payload is accepted, I'm not too sure actually. I'd say the API would through smth like an invalid address error. Either way whatever the merchant sends through our library, it's the API that should tell them what fields to populate :)

Adyen/Model/Payout/Address.cs
```diff
@@ -61,28 +61,28 @@ public Address(string city = default(string), string country = default(string),
...
         public string City { get; set; }
 
         /// <summary>
         /// The two-character ISO-3166-1 alpha-2 country code. For example, **US**. &gt; If you don&#39;t know the country or are not collecting the country from the shopper, provide &#x60;country&#x60; as &#x60;ZZ&#x60;.
```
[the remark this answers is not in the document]
jillingk: Hmm interesting, I'll look into it

AlexandrosMor (approved):

wboereboom (approved):

PR #747, opened 2023-02-22
[ITT-366] Generated Service/Model for BalanceControl and add small UT

jillingk:
**Description**
This PR generates Service/Model for BalanceControl and add small UT


Adyen/Model/BalanceControl/AbstractOpenAPISchema.cs
```diff
@@ -0,0 +1,78 @@
...
+* The version of the OpenAPI document: 1
+* Contact: developer-experience@adyen.com
+*
+* NOTE: This class is auto generated by OpenAPI Generator (https://openapi-generator.tech).
+* https://openapi-generator.tech
+* Do not edit the class manually.
+*/
+
+using System;
+using Newtonsoft.Json;
```
wboereboom, line 14: weren't we going to move away from Newtonsoft?

wboereboom (approved):

Adyen/Model/BalanceControl/AbstractOpenAPISchema.cs
jillingk: not anytime soon :(

Adyen/Model/BalanceControl/AbstractOpenAPISchema.cs
```diff
@@ -0,0 +1,78 @@
...
+                {
+                    OverrideSpecifiedNames = false
+                }
+            }
+        };
+
+        /// <summary>
+        ///  Custom JSON serializer for objects with additional properties
+        /// </summary>
+        static public readonly JsonSerializerSettings AdditionalPropertiesSerializerSettings = new JsonSerializerSettings
```
Kwok-he-Chu, line 44: This returns a JsonSerializerSettings object withMissingMemberHandling.Ignore, let's rename it accordingly. We can add in the summary that it's used for objects with additional properties.

Or even better: create a methodthat returns a JsonSerializerSetting. This method can have different parameters where you can also specifiy the value for MissingMemberHandling --> a factory can create the JsonSerializerSettings.

Kwok-he-Chu: same for:   `static public readonly JsonSerializerSettings SerializerSettings`

Kwok-he-Chu: Could you give me a brief example of how these JsonSerializerSettings are used? Which is used out of the two? SerializerSettings vs. AdditionalPropertiesSerializerSettings

jillingk: They only really get used for one-of models that we use. It basically just checks first if the deserializable json is an exact match to a model, and then if it is a match with additional properties (hence the MissingMemberHandling.Ignore). For this endpoint we don't have any one-of models though, so it doesn't do anything :)

Adyen.Test/BalanceControlTest.cs
```diff
@@ -0,0 +1,21 @@
...
+        [TestMethod]
+        public void BalanceTransferTest()
+        {
+            var client = CreateMockTestClientApiKeyBasedRequestAsync("Mocks/balance-control-transfer.json");
+            var service = new BalanceControlService(client);
+            var response = service.BalanceTransfer(new BalanceTransferRequest());
+            Assert.AreEqual(response.CreatedAt, new DateTimeOffset(new DateTime(2022,01, 24, 14,59, 11)));
+            Assert.AreEqual(response.Status, BalanceTransferResponse.StatusEnum.Transferred);
+        }
+    }it
```
AlexandrosMor, line 20:
```suggestion
    }
```

AlexandrosMor (approved):

antolo-arch (approved):

PR #761, opened 2023-03-08
Regenerate POSTerminalManagement and remove resource classes

jillingk:
**Description**
This PR regenerated PosTerminalManagement in preparation of the release. Additionally it removes some unused resource classes.


Makefile
```diff
@@ -58,7 +58,7 @@ $(services): target/spec $(openapi-generator-jar)
 	mv target/out/src/Adyen.Model/$@/* Adyen/Model/$@
 
 # Generate a full client (models and service classes)
-LegalEntityManagement: target/spec $(openapi-generator-jar)  
+PosTerminalManagement: target/spec $(openapi-generator-jar)
```
michaelpaul, line 61: We can introduce a variable here instead of replacing it for every service: 
```suggestion
full-services:=LegalEntityManagement PosTerminalManagement
$(full-services): target/spec $(openapi-generator-jar)  
```

AlexandrosMor (approved):

PR #776, opened 2023-03-27
[DR-182] Adjust templates to fix incorrect Long (de)serialization

jillingk:
**Description**
@Kwok-he-Chu ran into a problem with his example integration. After quickly looking at it I noticed we have a lot of new Long values in our library that run into the same serialization problem as we had with ints. I introduced the same fix for ints to fix this in the templates.


AlexandrosMor (approved):

AlexandrosMor (dismissed):

wboereboom (commented):
could we also use `isNumeric` here?

jillingk:
> isNumeric

Let me test it out

wboereboom (approved):

michaelpaul (approved):

michaelpaul (approved):

AlexandrosMor (approved):

PR #783, opened 2023-04-04
[ITT-201] Add Service Interfaces

jillingk:
**Description**
This PR adds interfaces to the services through the api mustache templates. To test it I regenerated checkout.

### DIFF
--- a/Adyen/Service/Checkout/ClassicCheckoutSDKService.cs
+++ b/Adyen/Service/Checkout/ClassicCheckoutSDKService.cs
@@ -25,7 +25,7 @@ namespace Adyen.Service.Checkout
     /// <summary>
     /// Represents a collection of functions to interact with the API endpoints
     /// </summary>
-    public class ClassicCheckoutSDKService : AbstractService
+    public class ClassicCheckoutSDKService : AbstractService, IClassicCheckoutSDKService
     {
         private readonly string _baseUrl;
         
@@ -34,27 +34,12 @@ namespace Adyen.Service.Checkout
             _baseUrl = client.Config.CheckoutEndpoint + "/" + ClientConfig.CheckoutApiVersion;
         }
     
-        /// <summary>
-        /// Create a payment session
-        /// </summary>
-        /// <param name="idempotencyKey">A unique identifier for the message with a maximum of 64 characters (we recommend a UUID).</param>
-        /// <param name="paymentSetupRequest"></param>
-        /// <param name="requestOptions">Additional request options.</param>
-        /// <returns>PaymentSetupResponse</returns>
         [Obsolete]
         public PaymentSetupResponse PaymentSession(PaymentSetupRequest paymentSetupRequest, RequestOptions requestOptions = default)
         {
             return PaymentSessionAsync(paymentSetupRequest, requestOptions).GetAwaiter().GetResult();
         }
 
-        /// <summary>
-        /// Create a payment session
-        /// </summary>
-        /// <param name="idempotencyKey">A unique identifier for the message with a maximum of 64 characters (we recommend a UUID).</param>
-        /// <param name="paymentSetupRequest"></param>
-        /// <param name="requestOptions">Additional request options.</param>
-        /// <param name="cancellationToken"> A CancellationToken enables cooperative cancellation between threads, thread pool work items, or Task objects.</param>
-        /// <returns>Task of PaymentSetupResponse</returns>
         [Obsolete]
         public async Task<PaymentSetupResponse> PaymentSessionAsync(PaymentSetupRequest paymentSetupRequest, RequestOptions requestOptions = default, CancellationToken cancellationToken = default)
         {
@@ -63,27 +48,12 @@ namespace Adyen.Service.Checkout
             return await resource.RequestAsync<PaymentSetupResponse>(paymentSetupRequest.ToJson(), requestOptions, new HttpMethod("POST"), cancellationToken);
         }
 
-        /// <summary>
-        /// Verify a payment result
-        /// </summary>
-        /// <param name="idempotencyKey">A unique identifier for the message with a maximum of 64 characters (we recommend a UUID).</param>
-        /// <param name="paymentVerificationRequest"></param>
-        /// <param name="requestOptions">Additional request options.</param>
-        /// <returns>PaymentVerificationResponse</returns>
         [Obsolete]
         public PaymentVerificationResponse VerifyPaymentResult(PaymentVerificationRequest paymentVerificationRequest, RequestOptions requestOptions = default)
         {
             return VerifyPaymentResultAsync(paymentVerificationRequest, requestOptions).GetAwaiter().GetResult();
         }
 
-        /// <summary>
-        /// Verify a payment result
-        /// </summary>
-        /// <param name="idempotencyKey">A unique identifier for the message with a maximum of 64 characters (we recommend a UUID).</param>
-        /// <param name="paymentVerificationRequest"></param>
-        /// <param name="requestOptions">Additional request options.</param>
-        /// <param name="cancellationToken"> A CancellationToken enables cooperative cancellation between threads, thread pool work items, or Task objects.</param>
-        /// <returns>Task of PaymentVerificationResponse</returns>
         [Obsolete]
         public async Task<PaymentVerificationResponse> VerifyPaymentResultAsync(PaymentVerificationRequest paymentVerificationRequest, RequestOptions requestOptions = default, CancellationToken cancellationToken = default)
         {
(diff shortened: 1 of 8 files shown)


templates/csharp/api.mustache
```diff
@@ -60,5 +55,33 @@ namespace {{packageName}}.{{apiPackage}}
         {{/supportsAsync}}
         {{/operation}}
     }
+    
     {{/operations}}
-}
+    {{#operations}}
```
Kwok-he-Chu, line 60: I'd love to have the interface in a separate class. The implementation might have other namespaces that the interface doesn't need to know about.

However, if you want to go forward with 1 single class, I'd put the interface on top and the implementation on the bottom 👍

Adyen/Service/Checkout/OrdersService.cs
```diff
@@ -34,18 +34,58 @@ public OrdersService(Client client) : base(client)
...
+    
+    /// <summary>
+    /// Service Interface
+    /// </summary>
+    public interface IOrdersService
+    {
         /// <summary>
         /// Create an order
         /// </summary>
         /// <param name="idempotencyKey">A unique identifier for the message with a maximum of 64 characters (we recommend a UUID).</param>
```
Kwok-he-Chu, line 83: I don't see a idempotencyKey here (this is probably moved to the Client level (just a guess*), we can safely remove this comment 👍

Adyen/Service/Checkout/OrdersService.cs
```diff
@@ -34,18 +34,58 @@ public OrdersService(Client client) : base(client)
...
         /// <param name="checkoutCreateOrderRequest"></param>
```
Kwok-he-Chu, line 84: - [ ] We should move the comments to Interface level only (that's what the developers will use/see)
- [ ] Point of improvement: We could annotate the class properl instead of leaving the param empty
```csharp
/// <param name="checkoutCreateOrderRequest"><see cref="CheckoutCreateOrderRequest"/>.</param>
```
- [ ] Point of improvement 2: We could do the same for the return type
```csharp
/// <returns ><see cref="CheckoutCreateOrderResponse"/>.</returns>
```

Kwok-he-Chu (changes requested):
Nice work, almost there 👍

Adyen/Service/Checkout/OrdersService.cs
Kwok-he-Chu: There are multiple functions/methods that have the same 'idempotencyKey'-comment ^ Not just this one*
Please remove it 👍

jillingk: yeah this is an annoying one tbh. Not really the scope of this PR but I'll see if I can remove it or if not create a ticket for it.

Adyen/Service/Checkout/OrdersService.cs
jillingk: I did move comments to interface, bit hard to see bc of this messed up git diff :)

jillingk:
@Kwok-he-Chu fixed most of your comments, will look into the idempotency tomorrow!

templates/csharp/api.mustache
wboereboom: Looking at the documentation it takes a substantial effort to have the generator create new files rather than modifying templates for existing ones. @jillingk maybe we can make a ticket to do this somewhere in the future as a nice-to-have?

templates/csharp/api.mustache
```diff
@@ -12,22 +12,50 @@ using Adyen.Service.Resource;
...
+        /// <returns><see cref="{{{returnType}}}"/>.</returns>
+        {{/returnType}}
+        {{#isDeprecated}}
+        [Obsolete]
+        {{/isDeprecated}}
+        {{{returnType}}}{{^returnType}}void{{/returnType}} {{#lambda.pascalcase}}{{vendorExtensions.x-methodName}}{{/lambda.pascalcase}}({{>api_parameters}});
+        
+        {{#supportsAsync}}
+{{>method_documentation}}
+        /// <param name="cancellationToken"> A CancellationToken enables cooperative cancellation between threads, thread pool work items, or Task objects.</param>{{#returnType}}
```
AlexandrosMor, line 33: Is this hard coded ?

Kwok-he-Chu: Correct, we should add the cancellationToken to our async methods on service-level and client-level. We have a ticket open to make that change as well 👍

Kwok-he-Chu:
Nice work so far @jillingk !

Adyen/Service/Checkout/RecurringService.cs
```diff
@@ -22,40 +22,70 @@
...
+    {
+        /// <summary>
+        /// Delete a token for stored payment details
+        /// </summary>
+        /// <param name="recurringId"><see cref="string"/> - The unique identifier of the token.</param>
+        /// <param name="shopperReference"><see cref="string"/> - Your reference to uniquely identify this shopper, for example user ID or account ID. Minimum length: 3 characters. &gt; Your reference must not include personally identifiable information (PII), for example name or email address.</param>
+        /// <param name="merchantAccount"><see cref="string"/> - Your merchant account.</param>
+        /// <param name="requestOptions"><see cref="RequestOptions"/> - Additional request options.</param>
+        /// <returns><see cref="StoredPaymentMethodResource"/>.</returns>
+        StoredPaymentMethodResource DeleteTokenForStoredPaymentDetails(string recurringId, string shopperReference = default, string merchantAccount = default, RequestOptions requestOptions = default);
```
Kwok-he-Chu, line 38: `recurringId`, I believe this called `recurringDetailReference` Let's rename this and keep it consistent.

src: https://docs.adyen.com/online-payments/classic-integrations/api-integration-ecommerce/recurring-payments/disable-stored-details#request

Adyen/Service/Checkout/RecurringService.cs
```diff
@@ -22,40 +22,70 @@
...
+        
+        /// <summary>
+        /// Delete a token for stored payment details
+        /// </summary>
+        /// <param name="recurringId"><see cref="string"/> - The unique identifier of the token.</param>
```
Kwok-he-Chu, line 43: `recurringDetailReference`

Adyen/Service/Checkout/RecurringService.cs
```diff
@@ -22,40 +22,70 @@
...
 {
+    /// <summary>
+    /// RecurringService Interface
+    /// </summary>
+    public interface IRecurringService
+    {
+        /// <summary>
+        /// Delete a token for stored payment details
+        /// </summary>
+        /// <param name="recurringId"><see cref="string"/> - The unique identifier of the token.</param>
```
Kwok-he-Chu, line 33: `recurringDetailReference`

Adyen/Service/Checkout/RecurringService.cs
jillingk: The link you sent is classic payments, plus it seems to be quite old (v25) so I think the specs here are right!

templates/csharp/api.mustache
```diff
@@ -12,22 +12,50 @@ using Adyen.Service.Resource;
...
+        {{/operation}}
+    }
+    {{/operations}}
+    
     {{#operations}}
     /// <summary>
     /// Represents a collection of functions to interact with the API endpoints
     /// </summary>
-    {{>visibility}} class {{classname}} : AbstractService
+    {{>visibility}} class {{classname}} : AbstractService, I{{classname}}
```
Kwok-he-Chu, line 49: @jillingk One last thing, please update the readme accordingly, there might be some services that are instantiated differently (also the names e.g. 'Checkout' in readme are incorrect) 👍

templates/csharp/api.mustache
AlexandrosMor: Cool thanks guys @Kwok-he-Chu @jillingk !

AlexandrosMor (dismissed):

templates/csharp/api.mustache
AlexandrosMor: @jillingk

jillingk: @Kwok-he-Chu, the Readme should be correct. We opted to define the directory for services with subservices (hence Checkout/ ) and if not the actual class name (eg StoredValueService). Old services that we didn't change, such as classic platforms and terminal API, I left as is. Does that makes sense to you, or would you rather see something else?

Kwok-he-Chu: I should've been more specific, sorry - let me clarify:

In https://github.com/Adyen/adyen-dotnet-api-library/blob/develop/README.md?plain=1#L48-L73 we create `new Checkout` which no longer exists and is pointed to IPaymentsService (developers should now create a new PaymentsService object instead) or even better: inject the services.AddSingleton<IPaymentsService, PaymentsService>() into their servicecollection container.

seealso https://github.com/Adyen/adyen-dotnet-api-library/blob/develop/Adyen/Service/Checkout/PaymentsService.cs
Note: The namespaces/references in the readme probably are incorrectly mentioned here


In https://github.com/Adyen/adyen-dotnet-api-library/blob/10.0.0-beta/README.md?plain=1#L11-L26, we should update the Service names (maybe even reference to have it link to the relative foldername in the readme, e.g. /checkout would go to the following folder: https://github.com/Adyen/adyen-dotnet-api-library/tree/10.0.0-beta/Adyen/Service/Checkout (You can have relative links by adding it as `[test](Adyen/Service/Checkout)` for example)

wboereboom: @Kwok-he-Chu should we also update the README examples?

jillingk: @Kwok-he-Chu, great feedback thanks. I changed it except for the IServiceCollection injection. I want to create a little codesnippet for a generic .NET webapp setup but for that I first want to play around with it a bit so I know what works ;)

michaelpaul (approved):

wboereboom (approved):

Adyen/Service/Checkout/ClassicCheckoutSDKService.cs
```diff
@@ -22,75 +22,89 @@
...
 {
+    /// <summary>
+    /// ClassicCheckoutSDKService Interface
+    /// </summary>
+    public interface IClassicCheckoutSDKService
+    {
+        /// <summary>
+        /// Create a payment session
+        /// </summary>
+        /// <param name="paymentSetupRequest"><see cref="PaymentSetupRequest"/> - </param>
```
Kwok-he-Chu, line 33: @jillingk `paymentRequest -` has no description (while e.g. requestOptoins does have a descriptions), checking whether this was intended. Final suggestion to just add leave it as a `><see cref="PaymentSetupRequest"/>.` since the class PaymentSetupRequest will have the description

Kwok-he-Chu (approved):

PR #788, opened 2023-04-08
[ITT-430] Generate entire library and fix tests and minor service maintenance

jillingk:
**Description**
This PR uses #786 templates and makefile to generate all (recent endpoints) services/models again.


AlexandrosMor (approved):

Adyen.IntegrationTest/BaseTest.cs
```diff
@@ -22,7 +22,7 @@ public class BaseTest
         public PaymentResult CreatePaymentResult()
         {
             var client = CreateApiKeyTestClient();
-            var payment = new PaymentService(client);
+            var payment = new GeneralService(client);
```
michaelpaul, line 25: `GeneralService` doesn't look better than the previous `PaymentService`. How about treating classic PAL as single file service and keep the old name?

jillingk: @michaelpaul, I was looking at doing it that way. However I ran into the problem of not being able to figure out how to squash two or more groups/tags into one file. If you have any idea please let me know

wboereboom (approved):

Issue #789, opened 2023-04-11
Due to non-usage of `ConfigureAwait(false)` the library hangs with .Net 48 Web API code 


bahrianonline:
**Describe the bug**
Due to non-usage of `ConfigureAwait(false)` the library hangs with .Net 48 Web API code. Ref: https://devblogs.microsoft.com/dotnet/configureawait-faq/

**To Reproduce**
As the code is called to /payments api (using PaymentsService. PaymentsAsync), it never returns. 
I have updated the code as follows and it fixed the issue.

`        public async Task<PaymentResponse> PaymentsAsync(PaymentRequest paymentRequest, RequestOptions requestOptions = null)
        {
            string endpoint = _baseUrl + "/payments";
            return await new ServiceResource(this, endpoint).RequestAsync<PaymentResponse>(paymentRequest.ToJson(), requestOptions, new HttpMethod("POST")).ConfigureAwait(continueOnCapturedContext: false);
        }`

**Expected behavior**
It shouldn't hang.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome
 - Version Latest

**Additional context**
I tried to create the pull request but don't have the access to do so.

bahrianonline:
Any update on this please?

jillingk:
Hi @bahrianonline,

We have thought about this and it we decided that we'd rather actually not set this continueOnCapturedContext flag to false by default in the ConfigureAwaiter. I definitely see the benefit of setting the default to false throughout the library, for improved performance, avoiding deadlocks etc. However since the .NET default is true, and this setting works in almost every case a developer would find himself in (albeit with some drawbacks), we do think this is the best solution. 

I know it might be a bit tedious to set this flag on every async call, but it prevents a lot of confusion I'm afraid. If you have any good arguments why we definitely should set it to false, we're always open te review this again though.

Best, Jilling
Adyen

bahrianonline:
@jillingk : Thanks for the update. But, as I have mentioned under .Net 4X it creates the hang due to that I had to update the library in my project and use it locally with ConfigureAwait(false). 

You can check this link https://blog.stephencleary.com/2017/03/aspnetcore-synchronization-context.html:

It states clearly about the issue I am facing, as we are using ASP.NET app and it hangs without it:

> **You Don’t Need ConfigureAwait(false), But Still Use It in Libraries**
> 
[...]

PR #806, opened 2023-05-09
[ITT-527] Add ConfigureAwait(false) flag to every async/await block

jillingk:
**Description**
This PR adds ConfigureAwait flags to every async/await block to prevent deadlocks and more for older .NET frameworks.


wboereboom (approved):

wboereboom:
this fixed #789

michaelpaul (approved):

PR #813, opened 2023-05-22
Fix model generation workflow and generate models

jillingk:
**Description**
fix model generation and generate models.

### DIFF
--- a/Adyen/Service/BalancePlatform/BankAccountValidationService.cs
+++ b/Adyen/Service/BalancePlatform/BankAccountValidationService.cs
@@ -42,7 +42,7 @@ namespace Adyen.Service.BalancePlatform
         /// <param name="requestOptions"><see cref="RequestOptions"/> - Additional request options.</param>
         /// <param name="cancellationToken"> A CancellationToken enables cooperative cancellation between threads, thread pool work items, or Task objects.</param>
         /// <returns>void</returns>
-        void ValidateBankAccountIdentificationAsync(BankAccountIdentificationValidationRequest bankAccountIdentificationValidationRequest, RequestOptions requestOptions = default, CancellationToken cancellationToken = default);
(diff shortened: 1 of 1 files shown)


Adyen/Service/BalancePlatform/BankAccountValidationService.cs
```diff
@@ -40,6 +41,7 @@ public interface IBankAccountValidationService
         /// <param name="bankAccountIdentificationValidationRequest"><see cref="BankAccountIdentificationValidationRequest"/> - </param>
         /// <param name="requestOptions"><see cref="RequestOptions"/> - Additional request options.</param>
         /// <param name="cancellationToken"> A CancellationToken enables cooperative cancellation between threads, thread pool work items, or Task objects.</param>
+        /// <returns>void</returns>
```
wboereboom, line 44: does not return void

Adyen/Service/LegalEntityManagement/DocumentsService.cs
```diff
@@ -33,7 +33,7 @@ public interface IDocumentsService
         /// <param name="id"><see cref="string"/> - The unique identifier of the document to be deleted.</param>
         /// <param name="requestOptions"><see cref="RequestOptions"/> - Additional request options.</param>
         /// <returns><see cref="Object"/>.</returns>
-        void DeleteDocument(string id, RequestOptions requestOptions = default);
+        Object DeleteDocument(string id, RequestOptions requestOptions = default);
```
wboereboom, line 36: Should this still be void?

wboereboom: There are m ore instances of this in this services class

wboereboom (changes requested):
Still some Object responses. see comments about the ones I found

PR #873, opened 2023-08-28
Fix/automation/services

wboereboom:
**Description**
This PR contains the automation changes + a fix for the unit tests due to some service changes.


jillingk (dismissed):

Adyen/Service/BalancePlatform/NetworkTokensService.cs
```diff
@@ -0,0 +1,99 @@
+/*
+* Configuration API
+*
+*
+* The version of the OpenAPI document: 2
```
Aleffio, line 5: do we need a row on the Readme for Configuration API on this library?
we can have something similar as 

https://github.com/Adyen/adyen-node-api-library/blob/c2e71abdacd90262d5e1310d3e0fbe27a12b6e1f/README.md?plain=1#L22

jillingk (approved):

PR #880, opened 2023-09-01
Fix cloud live api endpoints

jillingk:
**Description**
This PR fixes the issue with the cloudterminal api endpoint getting overwritten in the client. Additionally it also support setting the live api region by setting an enum in the config.

### DIFF
--- a/Adyen/Constants/ClientConfig.cs
+++ b/Adyen/Constants/ClientConfig.cs
@@ -8,6 +8,7 @@
         public static string CloudApiEndPointEULive = "https://terminal-api-live.adyen.com";
         public static string CloudApiEndPointAULive = "https://terminal-api-live-au.adyen.com";
         public static string CloudApiEndPointUSLive = "https://terminal-api-live-us.adyen.com";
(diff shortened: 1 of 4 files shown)


Adyen/Constants/ClientConfig.cs
```diff
@@ -8,6 +8,7 @@ public class ClientConfig
         public static string CloudApiEndPointEULive = "https://terminal-api-live.adyen.com";
         public static string CloudApiEndPointAULive = "https://terminal-api-live-au.adyen.com";
         public static string CloudApiEndPointUSLive = "https://terminal-api-live-us.adyen.com";
+        public static string CloudApiEndPointAPSELive = "https://terminal-api-live-apse.adyen.com";
```
wboereboom, line 11: General question: How do we expect our merchants to be aware of these public fields? How would they when to use them and how to find them?

jillingk: Good question, there is an example on the README. Then selecting ClientConfig shows them the different options. That should be enough right or?

wboereboom (approved):

michaelpaul (approved):

PR #916, opened 2023-10-12
Allow baseUrl to be overridden in a test environment

PG-RichT:
**Description**

This PR adds the following changes:

- Extends the Config class and introduces a new BaseUrlConfig class. Please advise if the location of this is suitable, it's currently in the project root with the Config class.
- Extends the AbstractService to allow the baseUrl to be overridden when in a test environment. 

This functionality is extremely useful for routing traffic via a mock server to return responses that are not easily reproduceable on the test environment.

**Tested scenarios**

Unit tests are passing. If the BaseUrlConfig object is not included when the Adyen Client is setup, then the new code has no affect on the existing flow. If the environment is set to Live, the new code has no affect on the existing flow.

**Fixed issue**:  
N/A

### DIFF
--- a/Adyen/BaseUrlConfig.cs
+++ b/Adyen/BaseUrlConfig.cs
@@ -1,0 +1,20 @@
+﻿namespace Adyen
+{
+    public class BaseUrlConfig
+    {
+        /// <summary>
+        /// Url to use for Checkout actions. Ensure no trailing slashes.
+        /// </summary>
+        public string CheckoutUrl { get; set; }
+        
+        /// <summary>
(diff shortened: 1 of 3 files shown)


PG-RichT:
@jillingk Thanks for responding, is there any update or a timeline for a decision on this approach?

Aleffio (approved):

jillingk (approved):

jillingk:
Hey @PG-RichT,

Thanks a lot for this contribution! We highly appreciate you taking the time to improve this library. Let us know if you run into anything else, or if anything is unclear during your integration.

Best, Jilling
Adyen

PR #922, opened 2023-10-23
Release v12.0.0

AdyenAutomationBot:
Merged pull requests to be released:


jsyrjala:
Adyen library v12.0.0 is already available in package repositories, even though according git you have not yet released that:
https://www.nuget.org/packages/Adyen/12.0.0

jillingk:
Hi @jsyrjala,

Thanks for reaching out here. We actually made a mistake during the release yesterday and have instead of releasing the latest package under a beta tag, released a major version of the library. Not too worry though as this release is still stable. The reason we wanted to do a beta release is because we're gonna do major changes again soon for the checkout API (as they're bumping the API version to v71) and but we did want to introduce the latest changes already to test it out. We will follow this release up with a v13 release very soon, which will be paired with a migration guide with the all the changes for the v11 -> v13 bump :)

Best, Jilling
Adyen

PR #954, opened 2023-12-06
Add region selection for cloud terminal API endpoints

AlexandrosMor:
**Description**

Currently we have a few constants containing cloud API urls, yet we have no selection mechanism for merchants to actually select the right region. In this pull request merchants could pass the region using the `Config` and `TerminalApiRegion` field. Default is `EU` as it was if no value is passed. 

**Tested scenarios**
Add test covering all the new region enums

### DIFF
--- a/Adyen/Client.cs
+++ b/Adyen/Client.cs
@@ -4,6 +4,7 @@ using Adyen.Constants;
 using Adyen.Exceptions;
 using Adyen.HttpClient;
 using Adyen.HttpClient.Interfaces;
+using Adyen.Service.Resource.Terminal;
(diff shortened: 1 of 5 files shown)


Adyen/Client.cs
```diff
@@ -80,12 +81,23 @@ public void SetEnvironment(Environment environment, string liveEndpointUrlPrefix
             {
                 return Config.CloudApiEndPoint;
             }
-
+
```
wboereboom, line 84:
```suggestion
```

wboereboom (approved):

Issue #943, opened 2023-11-27
API error when calling PlatformsNotificationConfigurationService.GetNotificationConfigurationListAsync


p-zender:
**Describe the bug**
`PlatformsNotificationConfigurationService.GetNotificationConfigurationListAsync` returns an error with a message "Empty input which would have resulted in a null result.". 

Based on what I heard from Adyen support this is because the library calls the API with a null request body instead of an empty JSON (and there is no way to send any request body that I'm aware of)

**To Reproduce**
1. Call `PlatformsNotificationConfigurationService.GetNotificationConfigurationListAsync `
2. An exception is thrown with the following information:

> 2023-11-24 12:25:25Z Error 400: BadRequest, ResponseBody: {"status":400,"errorCode":"702","message":"Empty input which would have resulted in a null result.","errorType":"validation"}
Adyen.HttpClient.HttpClientException: 400: BadRequest, ResponseBody: {"status":400,"errorCode":"702","message":"Empty input which would have resulted in a null result.","errorType":"validation"}
   at Adyen.HttpClient.HttpClientWrapper.RequestAsync(String endpoint, String requestBody, RequestOptions requestOptions, HttpMethod httpMethod, CancellationToken cancellationToken)
   at Adyen.Service.ServiceResource.RequestAsync[T](String json, RequestOptions requestOptions, HttpMethod httpMethod, CancellationToken cancellationToken)
   at Adyen.Service.PlatformsNotificationConfigurationService.GetNotificationConfigurationListAsync(RequestOptions requestOptions, CancellationToken cancellationToken)

**Expected behavior**
A list of currently configured notification subscriptions is returned

**Screenshots**

**Desktop (please complete the following information):**
 - N/A
 - N/A
 - N/A

**Smartphone (please complete the following information):**
 - N/A
 - N/A
 - N/A
 - N/A

**Additional context**
Using the 13.0.0 version of the client library.

wboereboom:
Hi @p-zender,

Thanks for raising this issue.
I've reproduced the issue on our end, and the library should indeed send in the empty body if the request object is empty.
We will fix this shortly.

However, I would like to point out that this endpoint is part of our Classic Platforms API, which we no longer support or update. I recommend reaching out to our support department to guide you through the upgrade process so you can use our new APIs and their features.

Kind Regards,
Wouter
Adyen

PR #971, opened 2024-01-11
Fix/empty post request

wboereboom:
**Description**
This fix ensures empty post requests will always get an empty Json string.  

**Tested scenarios**
Unit tests

**Fixed issue**:  #943

### DIFF
--- a/Adyen.Test/mockHttpClient.cs
+++ b/Adyen.Test/mockHttpClient.cs
@@ -1,0 +1,34 @@
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+public class MockHttpMessageHandler : HttpMessageHandler
+{
+    private readonly string _response;
+    private readonly HttpStatusCode _statusCode;
+
+    public string Input { get; private set; }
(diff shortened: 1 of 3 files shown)


Adyen.Test/mockHttpClient.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+public class MockHttpMessageHandler : HttpMessageHandler
+{
+    private readonly string _response;
+    private readonly HttpStatusCode _statusCode;
+
+    public string Input { get; private set; }
+    public int NumberOfCalls { get; private set; }
```
AlexandrosMor, line 12: Is the `NumberOfCalls` used ?

wboereboom: No, I used this mock based on an article on internet. If we want to include more HTTP client tests in the future we would be able to use this field. Or I can just remove it for now.

AlexandrosMor: It is a mock so it is ok

AlexandrosMor (approved):

PR #985, opened 2024-02-16
Support Checkout POST storedPaymentMethod

Lacutas82:
Add the ability to call POST /storedPaymentMethods

### DIFF
--- a/Adyen.Test/CheckoutTest.cs
+++ b/Adyen.Test/CheckoutTest.cs
@@ -1146,6 +1146,26 @@ namespace Adyen.Test
             var checkout = new MyRecurringService(client);
             checkout.DeleteTokenForStoredPaymentDetails("shopperRef", "merchantAccount");
         }
+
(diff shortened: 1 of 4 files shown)


wboereboom:
Hi @Lacutas82,

All our Service objects and models are added through code generation. This means any changes you propose would be overwritten. I'm forwarding your request to our internal Checkout team so they can investigating documenting this end-point. That means  it would get added in [our API explorer](https://docs.adyen.com/api-explorer) as well as in all our API Libraries if they decide this endpoint should indeed be documented as such. I'll keep this PR open to function as an issue tracker.

Kind Regards,
Wouter
Adyen

jillingk:
@Lacutas82 the endpoint has been added and we will release it later today!

Lacutas82:
> @Lacutas82 the endpoint has been added and we will release it later today!

Amazing @jillingk , thank you for the update!

Issue #995, opened 2024-03-20
Allow using encrypted payload with Cloud API


jsyrjala:
**Is your feature request related to a problem? Please describe.**

We'd like to use encrypted payloads with Cloud API. Especially we'd like to receive encrypted  async responses in Event notifications. And seems that responses would be encrypted if we would encrypt our requests. But currently SDK doesn't allow specifying encryption details.

With local api you can give `EncryptionCredentialDetails`
```csharp
public interface ITerminalLocalApi
  {
    SaleToPOIResponse TerminalRequest(
      SaleToPOIMessage saleToPoiRequest,
      EncryptionCredentialDetails encryptionCredentialDetails);

    Task<SaleToPOIResponse> TerminalRequestAsync(
      SaleToPOIMessage saleToPoiRequest,
      EncryptionCredentialDetails encryptionCredentialDetails,
      CancellationToken cancellationToken = default (CancellationToken));

    SaleToPOIResponse TerminalRequest(
      SaleToPOIMessage saleToPoiRequest,
      EncryptionCredentialDetails encryptionCredentialDetails,
      RemoteCertificateValidationCallback remoteCertificateValidationCallback);

    string DecryptNotification(
      string notification,
      EncryptionCredentialDetails encryptionCredentialDetails);
  }
```

With cloud api there is no way to give `EncryptionCredentialDetails`
```csharp
  public interface ITerminalCloudApi
  {
    SaleToPOIResponse TerminalRequestAsync(SaleToPOIMessage saleToPoiRequest);

    SaleToPOIResponse TerminalRequestSync(SaleToPOIMessage saleToPoiRequest);

    Task<SaleToPOIResponse> TerminalRequestAsynchronousAsync(
      SaleToPOIMessage saleToPoiRequest,
      CancellationToken cancellationToken = default (CancellationToken));

    Task<SaleToPOIResponse> TerminalRequestSynchronousAsync(
      SaleToPOIMessage saleToPoiRequest,
      CancellationToken cancellationToken = default (CancellationToken));
  }
```

**Describe the solution you'd like**

I'd like be able to specify encryption also with Cloud API

**Describe alternatives you've considered**
- Copying large parts of SDK and making it support encryption in Cloud API sound rather lot of work.
- Not having encryption in responses doesn't sound too secure.

Kwok-he-Chu:
Hi @jsyrjala 

- With **Terminal API local communications** your POS app communicates directly with the payment terminal over your local network.

[...]

PR #1060, opened 2024-09-13
Allow .NET Library to send encrypted payloads to the `/sync` and `/async` Cloud Terminal API endpoints

Kwok-he-Chu:
**Description**
Allows developers to send encrypted payloads using the Terminal Cloud API (`/sync` & `/async` endpoints)

**Fixed issue**:
#995

**Changes**:
* Added:
  * TerminalApiAsyncService -> Allows sending unencrypted and encrypted requests to `/async`-endpoint
  * TerminalApiSyncService -> Allows sending unencrypted and encrypted requests to `/sync`-endpoint
  * TerminalApiLocalService -> Allows sending unencrypted and encrypted requests to a custom endpoint (e.g. `https://198.51.100.1:8443/nexo`)
* `Adyen/Service/Resource/Terminal/TerminalApi.cs` has been removed/refactored into:
  * TerminalApi**Local**Client -> TerminalApi**Local**Service
  * TerminalApi**Sync**Client -> TerminalApi**Sync**Service
  * TerminalApi**Async**Client -> TerminalApi**Async**Service
  * Updated `README.md` with the new code snippets to reflect these changes
* Marked `Adyen/Service/TerminalCloudApi.cs` as `[Obsolete]`, developers should use one of the three service-classes above instead
  * Reason for change: The following `TerminalRequestAsynchronousAsync(...)`, `TerminalRequestSynchronousAsync(...)`, `TerminalRequestAsync(...)`, `TerminalRequestSync(...)` were confusing because of the name of the endpoints, I've split the two concerns: `/async` & `/sync` into the respective services

**Fixes:**
* BaseTest returns the mocked response when `RequestAsync(...)` is called, previously it only would return the response when `Request(...)` is called
* `Adyen/ApiSerialization/SaleToPoiMessageSecuredSerializer.cs` is now marked `public`, developers can use this class when needed

**Breaking change:**
- None, I've marked previous classes as `[Obsolete]` and mentioned the name of the new to the new class


Adyen/ApiSerialization/SaleToPoiMessageSecuredSerializer.cs
```diff
@@ -7,14 +7,14 @@
 
 namespace Adyen.ApiSerialization
 {
-    internal class SaleToPoiMessageSecuredSerializer
+    public class SaleToPoiMessageSecuredSerializer
     {
-        internal string Serialize(SaleToPoiMessageSecured saleToPoiMessageSecured)
+        public string Serialize(SaleToPoiMessageSecured saleToPoiMessageSecured)
```
jillingk, line 12: Good one. I believe there was a merchant requesting this to be public, and I don't see a reason why not too so good move here.

Adyen/Service/TerminalApiAsyncService.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// <summary>
+        /// Used to decrypt the notification received.
+        /// </summary>
+        /// <param name="notification">Notification.</param>
+        /// <param name="encryptionCredentialDetails"><see cref="EncryptionCredentialDetails"/>.</param>
+        /// <returns>Decrypted notification.</returns>
+        string DecryptNotification(string notification, EncryptionCredentialDetails encryptionCredentialDetails);
+    }
+
+    public class TerminalApiAsyncService : AbstractService, ITerminalApiAsyncService
```
jillingk, line 43: Why not specify this is the cloud service in the name? ie
```suggestion
    public class TerminalApiCloudAsyncService : AbstractService, ITerminalApiCloudAsyncService
```

jillingk:
Also for most services in the lib we provide a non async method by awaiting the async methods task. It's a bit trivial for most .NET developers but maybe nice to have.

Adyen/Service/TerminalApiAsyncService.cs
Kwok-he-Chu: I ensured that the URL endpoints match the service name for consistency in the library. I noticed that the library keeps the same naming-conventions for i.e. `Payments`-Api, or `Modifications`. In this case, I prefer to use the `terminal-api-` + `/sync` & `/async` suffixes respectively.


image

Kwok-he-Chu (commented):
> @jillingk Also for most services in the lib we provide a non async method by awaiting the async methods task. It's a bit trivial for most .NET developers but maybe nice to have.

I'll add the functionality and add the `HttpClient.Request(...)` synchronous calls as well


 @jillingk @wboereboom I noticed that with these changes, we'd be breaking the library. Should I can reintroduce the previous name of the class, mark it as `[Obsolete]` & point it to the new function? That way, we don't introduce a breaking change, what do you think?

Adyen/Service/TerminalApiAsyncService.cs
jillingk: Right right, well since the local doesn't have async/sync (correct me if I'm wrong) that should be clear enough

jillingk:
> @jillingk @wboereboom I noticed that with these changes, we'd be breaking the library. Should I can reintroduce the previous name of the class, mark it as `[Obsolete]` & point it to the new function? That way, we don't introduce a breaking change, what do you think?

Hmm yeah we just did a major. Maybe good to make it non breaking for now and add the tag that we will remove the old stuff next major release

wboereboom:
I'm also all for a Deprecation strategy with a softer landing. Might be good to already plan when you actually want to remove the old stuff. We could also communicate this to our community so they know what to expect.

Kwok-he-Chu:
I'll modify the PR and apply a non-breaking change fix next tuesday

Adyen/Service/TerminalApiAsyncService.cs
```diff
@@ -0,0 +1,124 @@
...
+        private readonly SaleToPoiMessageSerializer _saleToPoiMessageSerializer;
+        private readonly SaleToPoiMessageSecuredEncryptor _saleToPoiMessageSecuredEncryptor;
+        private readonly SaleToPoiMessageSecuredSerializer _saleToPoiMessageSecuredSerializer;
+        
+        public TerminalApiAsyncService(Client client) : base(client)
+        {
+            _saleToPoiMessageSerializer = new SaleToPoiMessageSerializer();
+            _saleToPoiMessageSecuredEncryptor = new SaleToPoiMessageSecuredEncryptor();
+            _saleToPoiMessageSecuredSerializer = new SaleToPoiMessageSecuredSerializer();
+            _asyncClient = new TerminalApiAsyncClient(this);
```
jillingk, line 69: The merchant will still have to setup a certificate here for the unencrypted flow. The `TerminalLocalApiUnencrypted ` bypasses the `ServerCertificateCustomValidationCallback` of the httphandler to not have to do this. Makes testing a lot easier. Maybe nice to also implement that here (for the unencrypted methods)

Kwok-he-Chu: I like the idea, however, after digging deeper, I noticed it can cause unintended behavior, see example below:

For unencrypted message sending, developers have to override the behavior manually on `HttpClient`-level. The .NET library currently creates a [new Client](https://github.com/Adyen/adyen-dotnet-api-library/blob/v20.0.0/Adyen/Service/TerminalLocalApiUnencrypted.cs#L27). This means that it will override any client settings (timeouts, other settings, baseUrls). I prefer the approach of allowing developers to pass their own HttpClient and be in control of the parameters they pass.

Kwok-he-Chu: Seealso sonar/code scanning: https://github.com/Adyen/adyen-dotnet-api-library/security/code-scanning/4

jillingk: Yeah the sonar scan is a standard message, certificates normally should be verified.

Creating a client, and overriding the existing one is not ideal, however this is not production code. We have got a lot of questions in the past about setting up encryption and certificates hence wanting to make the actual testing part as easy and gradual as possible. You can try to reuse the existing client and override the handler, however I don't think that's necessary as its test code (and cannot make unencrypted calls on live anyways).

Kwok-he-Chu: I've added documentation in the method `<summary>` to showcase how developers can override the `HttpClientHandler` behavior and pass it to the client

jillingk: That works as well

jillingk (approved):

DjoykeAbyah (approved):

PR #1062, opened 2024-09-30
Remove extra backslash for possdk

jillingk:
**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Adyen/Service/AbstractService.cs
+++ b/Adyen/Service/AbstractService.cs
@@ -96,7 +96,6 @@ namespace Adyen.Service
             // Change base url for Live environment
             if (url.Contains(PaymentPrefix))
             {
-                // TODO these errors are not easy catchable for some reason, should be fixed first
                 if (config.LiveEndpointUrlPrefix == default)
                 {
                     throw new InvalidOperationException(ExceptionMessages.MissingLiveEndpointUrlPrefix);
@@ -111,10 +110,19 @@ namespace Adyen.Service
                 {
                     throw new InvalidOperationException(ExceptionMessages.MissingLiveEndpointUrlPrefix);
                 }
-                
-                url = url.Replace("https://checkout-test.adyen.com/",
-                    "https://" + config.LiveEndpointUrlPrefix + "-checkout-live.adyenpayments.com/checkout/");
+
+                if (url.Contains("possdk"))
+                {
+                    url = url.Replace("https://checkout-test.adyen.com/",
+                        "https://" + config.LiveEndpointUrlPrefix + "-checkout-live.adyenpayments.com/");   
+                }
+                else
+                {
+                    url = url.Replace("https://checkout-test.adyen.com/",
+                        "https://" + config.LiveEndpointUrlPrefix + "-checkout-live.adyenpayments.com/checkout/");
+                }
             }
+                
                 
             // If no prefix is required just replace "test" -> "live"
             url = url.Replace("-test", "-live");
--- a/Adyen.Test/PosMobileTest.cs
+++ b/Adyen.Test/PosMobileTest.cs
@@ -1,0 +1,28 @@
+using System;
+using System.Net.Http;
+using Adyen.Model.PosMobile;
+using Adyen.Service;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using NSubstitute;
+using Environment = Adyen.Model.Environment;
+
+namespace Adyen.Test
+{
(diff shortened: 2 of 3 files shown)


Adyen/Service/AbstractService.cs
```diff
@@ -111,11 +110,20 @@ private static string ConstructLiveUrl(Config config, string url)
                 {
                     throw new InvalidOperationException(ExceptionMessages.MissingLiveEndpointUrlPrefix);
                 }
-                
-                url = url.Replace("https://checkout-test.adyen.com/",
-                    "https://" + config.LiveEndpointUrlPrefix + "-checkout-live.adyenpayments.com/checkout/");
+
+                if (url.Contains("possdk"))
```
Kwok-he-Chu, line 114: Suggestion: Can we move the full url to the service-level class of the possdk? Preferably, when instantiating the possdk-service, let's write-out the full live and test urls

jillingk: In that case we'd need to retrieve it from the specs and they only contain test urls with the exception of one spec. Would be a great initiative to add both the test and live urls to specs though. That'd make these methods redundant.

Kwok-he-Chu: I dug a bit deeper, I think the library introduced a bug, let's add the tests here moving forward:

The possdk is used in the `/sessions` endpoint of the POS Mobile endpoint (v68) (see [API Explorer](https://docs.adyen.com/api-explorer/possdk/68/post/sessions))

1. Looking at the [PosMobileService](https://github.com/Adyen/adyen-dotnet-api-library/blob/v20.0.0/Adyen/Service/PosMobileService.cs#L55), the endpoint that it points to is: 
   - `https://checkout-test.adyen.com/checkout/possdk/v68` 

2. In any service that in `/Checkout` f.e. [PaymentsService](https://github.com/Adyen/adyen-dotnet-api-library/blob/v20.0.0/Adyen/Service/Checkout/PaymentsService.cs#L142), the endpoint points to: `https://checkout-test.adyen.com/v71`

3. Hence why the following `/checkout/` gets appended twice, because as it matches the `https://checkout-test.adyen.com/` url and replaces it.
```csharp
url = url.Replace("https://checkout-test.adyen.com/", "https://" + config.LiveEndpointUrlPrefix + "-checkout-live.adyenpayments.com/checkout/");
```
  - E.g. `https://checkout-test.adyen.com/checkout/possdk/v68` becomes `https://{liveprefix}-checkout-live.adyen.com/checkout/checkout/`

My suggestion: let's not use these string-replacements & fix it on service-level where it also appends the liveprefix in the next update

jillingk: So this off case is not because of a bug in the library but because the people who create the possdk resource on main decided to go for a base url that does not fit the usual standard that checkout uses. I reached out to the dev that did this and it was indeed a mistake but since they're going to replace this service anyways there's no point in fixing this. What I'm doing here is just a temporary quick fix so it atleast works on our side.

What do you mean with fix it on service-level?

Kwok-he-Chu: > What I'm doing here is just a temporary quick fix so it atleast works on our side.
What would the future change look like? Are they changing the URL? 

Fixing it on service-level refers to, moving the urls to the "-Service"-class -> [PosMobileService.cs](https://github.com/Adyen/adyen-dotnet-api-library/blob/v20.0.0/Adyen/Service/PosMobileService.cs#L55) would f.e. contain the test url and the live url.


* [ ] Can you add the tests for the serices in the /Checkout folder and PosMobileService on how the live urls should look like?

jillingk: How do you want to put the live url in the service class? You mean instead of one `_baseUrl` param, have two parameters one for test and one for live?

Kwok-he-Chu: Yes, construct the url and pass it, example: `_baseUrlTest` /  `_baseUrlLive`

jillingk: I mean technically it is constructed once. Your `Client` is either configured for live or for test and you construct any service object with this. If you switch to live you will need to pass a client with live env setup which is done on constructing the service. I'm fine with having it explicitly set on the service but that's a little bit outside of the scope of this quick fix (involves changing template and rethinking the whole logic of creating base urls)

DjoykeAbyah (approved):

PR #1068, opened 2024-10-08
Fix mustache template: Only add the deprecation message to the Obsolete-tag if it exists

Kwok-he-Chu:
**Description**
This PR addresses the `[Obsolete]`-tag in the model and service mustache templates. These should only be shown if the message is available in the specs. 

**Tested scenarios**

Example with `CardDetails.cupsecureplusSmscode` 

**Before:**
```csharp
[Obsolete("Deprecated since Adyen Checkout API v -")]
```

**After:**
```csharp
[Obsolete("")]
```
_I kept the template easy-to-read, I didn't want to make an exception for the empty case, e.g. just`[Obsolete]` without the double quotes and brackets `("")`_

___

**Example with just the deprecation version:**
```csharp
[Obsolete("Deprecated since Adyen Checkout API v37.")]
        Model.Checkout.PaymentSetupResponse PaymentSession(PaymentSetupRequest paymentSetupRequest = default, RequestOptions requestOptions = default);
```

___

**Example with the deprecation message and version:**
```csharp
[Obsolete("Deprecated since Adyen Checkout API v49. Use `storedPaymentMethodId` instead.")]
        public string RecurringDetailReference { get; set; }
```

### DIFF
--- a/templates/csharp/api-single.mustache
+++ b/templates/csharp/api-single.mustache
@@ -23,7 +23,7 @@ namespace {{packageName}}.Service
         /// <returns><see cref="{{{returnType}}}"/>.</returns>
         {{/returnType}}
         {{#isDeprecated}}
-        [Obsolete]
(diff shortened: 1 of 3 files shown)


gcatanese (changes requested):
Changes are fine.

Keep into consideration that for the endpoints there is the `x-deprecatedInVersion` extension without `vendorExtensions.x-deprecatedMessage`, in this case the formatting could look wrong.
Suggestions:

- check also if `x-deprecatedMessage` is available
- format the message differently, instead of using `-` use a `.`

gcatanese (approved):
Changes are good

If you like check the changes in Java app where I have updated the templates to check separately for the 2 extensions (without assuming `message` is provided when `version` is found)

https://github.com/Adyen/adyen-java-api-library/pull/1367/commits/be75d2df81af80787adafb92dc6792125a319a01

PR #1069, opened 2024-10-08
Update all services

AdyenAutomationBot:
OpenAPI spec or templates produced changes on 08-10-2024   by [commit](https://github.com/Adyen/adyen-openapi/commit/586e99dd4c4ea83c91e0108acd0ab7e720c41b6e).


Adyen/Model/BalancePlatform/AccountHolder.cs
```diff
@@ -116,7 +116,7 @@ protected AccountHolder() { }
         /// Gets or Sets ContactDetails
         /// </summary>
         [DataMember(Name = "contactDetails", EmitDefaultValue = false)]
-        [Obsolete]
+        [Obsolete("Deprecated since Configuration API v - ")]
```
jillingk, line 119: @Kwok-he-Chu If there's no deprecated since maybe we should format this differently? Alternatively we can try to enforce this annotation on main if we think this is crucial enough

Kwok-he-Chu: Good catch. I have turned this into `[Obsolete("")]` for now (see [PR](https://github.com/Adyen/adyen-dotnet-api-library/pull/1068/files)), we'll enforce the specs to have the messages added from here on out (fix the problem at root-level)

Kwok-he-Chu (approved):

Adyen/Model/BalancePlatform/AccountHolder.cs
Kwok-he-Chu: Fixed

PR #1037, opened 2024-07-05
Update all services

AdyenAutomationBot:
OpenAPI spec or templates produced changes on 17-07-2024   by [commit](https://github.com/Adyen/adyen-openapi/commit/5472ebda33cff062e1f6342fdc10ffeab4e6adec).

PR #1031, opened 2024-06-25
Update all services

AdyenAutomationBot:
OpenAPI spec or templates produced changes on 03-07-2024   by [commit](https://github.com/Adyen/adyen-openapi/commit/31a56103e6baea7401a968bdd9869e8b5b56600f).

PR #1026, opened 2024-06-05
Update all services

AdyenAutomationBot:
OpenAPI spec or templates produced changes on 21-06-2024   by [commit](https://github.com/Adyen/adyen-openapi/commit/ec41f6754446125d3262c9df4020bea7724c8956).

PR #1070, opened 2024-10-08
updated readme with disclaimer on Enum values

DjoykeAbyah:
### Description
This PR updates the README to include a disclaimer regarding the potential for enum values to change in future releases. This change ensures that users are aware that they should not rely on the enum values being static, and should handle them dynamically or with caution in their implementations.

- Added a new section in the README under `## Disclaimer on Enum Values`.
- Explained that enum values may be modified, added, or removed in future releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,12 @@ or install it from package manager
 ```
 PM> Install-Package Adyen -Version x.x.x
 ```
+## Disclaimer on Enum Values
+
+Please note that the enum values used in this library may change in future releases. These values are not guaranteed to remain static, and they may be modified, added, or removed as the library evolves. We advise against hard-coding or relying on specific enum integer values in your implementation, as they may break compatibility with future versions of this library.
+
+We recommend regularly checking the latest version of the project for any updates to the enums and other critical components.
+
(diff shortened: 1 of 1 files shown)


Kwok-he-Chu (approved):

Kwok-he-Chu:
I've found several cases (but not limited to these* only), this will become a problem if the underlying integer value is stored instead of the respective `string`-value.
* https://github.com/Adyen/adyen-dotnet-api-library/pull/1037/files#diff-654e173618021475129caec200e9fd730e8f1ae2b5ee59966779fec564b28835L64
* https://github.com/Adyen/adyen-dotnet-api-library/pull/1031/files#diff-923be669a9c28428736a92f8f172cb98a48a765df4992204f74b4ecd4443fb2aR226
* https://github.com/Adyen/adyen-dotnet-api-library/pull/1026/files#diff-44bbf43931d61abce260295e53f7988d0b32d407dde99e3c877b39ec56a3036bR192

I'll open a discussion. As of currently, adding one (alphabetical-ordered) **enum** in the OpenApi specifications results in an immediate breaking change down-stream.

Issue #1093, opened 2024-12-13
SaleToPOIRequest serialization issue


alexander-olesiyuk-apaleo:
**Describe the bug**
Call to https://terminal-api-test.adyen.com/sync ends with `Could not parse header` message if default json naming strategy is overridden. `SaleToPoiMessageConverter` does not specify `ContractResolver` property of `serializer` explicitly. So, default resolver is used. But running app can change default resolver/naming strategy which can cause changes in property names after serialization. For example, `CamelCaseNamingStrategy` serializes message header as
```javascript
{
    "messageClass": "Service",
    "messageCategory": "Payment",
    "messageType": "Request",
    "serviceID": "<value>",
    "saleID": "<value>",
    "poiid": "<value>",
    "protocolVersion": "3.0"
}
``` 
As a result, server can not parse it.

**To Reproduce**
1. Update default naming strategy
```c#
JsonConvert.DefaultSettings = () => new JsonSerializerSettings
{
    ContractResolver = new DefaultContractResolver()
    {
        NamingStrategy = new CamelCaseNamingStrategy()
    }
};
``` 
2. Create request
```c#
new SaleToPOIRequest()
{
    MessageHeader = new MessageHeader
    {
        MessageType = MessageType.Request,
        MessageClass = MessageClassType.Service,
        MessageCategory = MessageCategoryType.Payment,
        SaleID = "<value>",
        POIID = "<value>",
        ServiceID = "<value>"
    },
    MessagePayload = null
}
``` 
3. Send request by calling `TerminalApiSyncService.RequestAsync`
4. It returns event notification with details `message=Could+not+parse+header`
```javascript
{
    "SaleToPOIRequest": {
        "MessageHeader": {
            "DeviceID": "<value>",
            "MessageCategory": "Event",
            "MessageClass": "Event",
            "MessageType": "Notification",
            "POIID": "N/A",
            "ProtocolVersion": "3.0",
            "SaleID": "N/A"
        },
        "EventNotification": {
            "RejectedMessage": "<value>",
            "EventToNotify": "Reject",
            "TimeStamp": "2024-12-13T14:50:10.233Z",
            "EventDetails": "message=Could+not+parse+header"
        }
    }
}
```

Kwok-he-Chu:
Fixed in #1095 Thank you for your contribution!

PR #1095, opened 2024-12-16
Explicitly configure ContractResolver for SaleToPOIMessage serialization

alexander-olesiyuk-apaleo:
**Description**
Explicitly configure `ContractResolver` for `SaleToPOIMessage` and `SaleToPoiMessageSecured` serialization to use `DefaultContractResolver`. This is necessary to always use the same property naming strategy and not depend on `JsonConvert.DefaultSettings` as it can be changed by running app which can cause failure when parsing request on the server.

**Tested scenarios**
Added tests for `SaleToPoiMessage` and `SaleToPoiMessageSecured` which compare results of serialization when `JsonConvert.DefaultSettings` are not set and configured to use `CamelCaseNamingStrategy`. For `SaleToPoiMessageSecured` cleared `SecurityTrailer.Nonce` and `NexoBlob` as they are different for each execution.

**Fixed issue**:  #1093

### DIFF
--- a/Adyen.Test/SerializerTest.cs
+++ b/Adyen.Test/SerializerTest.cs
@@ -1,12 +1,12 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Text;
 using Adyen.ApiSerialization;
 using Adyen.Model.Checkout;
 using Adyen.Model.TerminalApi;
+using Adyen.Security;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Serialization;
 using PaymentResponse = Adyen.Model.TerminalApi.PaymentResponse;
 
 namespace Adyen.Test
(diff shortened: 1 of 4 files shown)


JonasBr68 (approved):
Exactly what was need!

Adyen/ApiSerialization/Converter/SaleToPoiMessageSecuredConverter.cs
```diff
@@ -14,8 +14,6 @@ internal class SaleToPoiMessageSecuredConverter : JsonConverter
 
         public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
         {
-            serializer.NullValueHandling = NullValueHandling.Ignore;
```
Kwok-he-Chu, line 17: See comment above, let's introduce a different function instead so that existing integrations don't break.

Adyen/ApiSerialization/Converter/JsonConvertSerializerWrapper.cs
```diff
@@ -11,19 +13,31 @@ internal class JsonConvertSerializerWrapper
...
         internal static string Serialize(SaleToPoiMessageSecured saleToPoiMessageSecured)
         {
             return JsonConvert.SerializeObject(saleToPoiMessageSecured,
-                                               new SaleToPoiMessageSecuredConverter(),
-                                               new StringEnumConverter(),
-                                               new IsoDateTimeConverter { DateTimeFormat = DateTimeFormat });
+                GetSerializerSettings(new SaleToPoiMessageSecuredConverter()));
+        }
+
+        private static JsonSerializerSettings GetSerializerSettings(JsonConverter messageConverter)
```
Kwok-he-Chu, line 27: nit: Let's group everything in one function and remove the `GetSerializerSettings(...)` function. When navigating through the library, it only takes one jump to understand how the `JsonSerializerSettings` is constructed

Adyen.Test/SerializerTest.cs
```diff
@@ -87,9 +87,91 @@ public void CheckoutSessionResponseCheckForIdTest()
...
+
+            try
+            {
+                var serializedWithUpdatedDefaultSettings = GetSerializedSaleToPoiMessage(saleToPoiMessage);
+
+                Assert.AreEqual(serializedWithEmptyDefaultSettings, serializedWithUpdatedDefaultSettings);
+            }
+            finally
+            {
+                JsonConvert.DefaultSettings = null;
```
Kwok-he-Chu, line 112: Do we need need to set this to `null` afterwards? I think the the try/finally is redundant.

Kwok-he-Chu (changes requested):
Thank you for the contribution, I've left some minor comments.

Adyen/ApiSerialization/Converter/JsonConvertSerializerWrapper.cs
alexander-olesiyuk-apaleo: Does this work?
Or should we inline method and duplicate code which creates `JsonSerializerSettings`?

Adyen/ApiSerialization/Converter/SaleToPoiMessageConverter.cs
```diff
@@ -8,8 +8,6 @@ internal class SaleToPoiMessageConverter : JsonConverter
     {
         public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
         {
-            serializer.NullValueHandling = NullValueHandling.Ignore;
```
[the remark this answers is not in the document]
alexander-olesiyuk-apaleo: To be honest, I don't understand how this can help to fix the issue. `WriteJson` is abstaract method of `Newtonsoft.Json.JsonConverter` class which is called by `Newtonsoft.Json` internally when we call `JsonConvert.SerializeObject`. If we create new method with different name it won't be called by `Newtonsoft.Json`.
Also, I don't see how this can break existing integrations. These two converters (`SaleToPoiMessageConverter` and `SaleToPoiMessageSecuredConverter`) are internal and they are used only in `JsonConvertSerializerWrapper` where values for `NullValueHandling` and `MissingMemberHandling` are explicitly provided when `JsonSerializerSettings` is created.

Adyen/ApiSerialization/Converter/SaleToPoiMessageSecuredConverter.cs
alexander-olesiyuk-apaleo: The same as for `SaleToPoiMessageConverter`.

Adyen/ApiSerialization/Converter/SaleToPoiMessageConverter.cs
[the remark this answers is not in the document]
Kwok-he-Chu: Ah, you're right. I didn't see that these functions are inherited from ` : JsonConverter` - Ignore my proposed solution.

As for the breaking part, I initially noticed that the proposed change would change the `NullValueHandling` from `Ignore` (default: `Include`), this will then change the behavior when either the `SaleToPoiMessageConverter.WriteJson(...)`/`SaleToPoiMessageSecuredConverter.WriteJson(...)`-functions are called.

However, I've overlooked that you've moved these fields to the `GetSerializerSettings(...)`-function.

All is good! 👍

Adyen/ApiSerialization/Converter/JsonConvertSerializerWrapper.cs
Kwok-he-Chu: I'd consider initializing the JsonSerializerSettings once, as now it's created every function-call

alexander-olesiyuk-apaleo: Two private static fields with settings for `SaleToPOIMessage` and for `SaleToPoiMessageSecured`?

Adyen.Test/SerializerTest.cs
alexander-olesiyuk-apaleo: Splitted.

Adyen/ApiSerialization/Converter/JsonConvertSerializerWrapper.cs
Kwok-he-Chu: Works for me!

alexander-olesiyuk-apaleo: I'm afraid I haven't got an idea.

When `CreateSerializerSettings` was initially defined it created `JsonSerializerSettings` each time it is called. `CreateSerializerSettings` was called from both `Serialize` methods. So, `JsonSerializerSettings` was instantiated per each `Serialize` call (the same as was before for converters). We could move it to both `Serialize` methods but nothing would change in such case from settings instantiation standpoint (but would require to change configuration if two places if necessary). Also, we can make settings static not to instantiate it per each call which is done in current implementation. Is there any other approach? Am I missing something?

Could you explain more clearly which solution you suggest, please?

Kwok-he-Chu: I was referring to: We could pass the `JsonSerializerSettings` in the parameters of the `Serialize(SaleToPoiMessageSecured saleToPoiMessageSecured, JsonSerializerSettings ...)` and `Serialize(SaleToPOIMessage saleToPoiMessage, JsonSerializerSettings ...)` methods. This allows consumers of the library to pass their (custom) `JsonSerializerSettings` if needed.

I like your current approach, this will do fine.

Kwok-he-Chu (approved):

Adyen/ApiSerialization/Converter/JsonConvertSerializerWrapper.cs
alexander-olesiyuk-apaleo: I don't think it is worth doing (pass `JsonSerializerSettings` to `Serialize` methods), at lease now, for two reasons:
1. Passing custom settings in most cases will generate json which will be refused by the server.
2. Settings are used to serialize header and payload objects but some other things like top object and header and payload property names are serialized by writer (e.g. `writer.WritePropertyName(saleToPoiMessage.MessageHeader.GetType().Name)`) where settings are ignored, as I understand.

alexander-olesiyuk-apaleo:
@Kwok-he-Chu, could you merge my changes, please? I don't have permissions even after your approval. Thank you.

Kwok-he-Chu:
Merged, thank you for the contribution! @alexander-olesiyuk-apaleo

Issue #1131, opened 2025-04-05
[Feature]: .NET 8 Support


fbiel:
### Feature summary

Support for .NET 8 (LTS) – .NET 6 is End-of-Life

### Problem statement

The SDK currently only supports .NET 6, which reached [End of Life on November 12, 2024](https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-and-net-core). This creates a significant security liability, as there is no overlap between supported .NET versions and the SDK’s compatibility.

The latest .NET Long-Term Support version is .NET 8, released on November 14, 2023. Support for .NET 8 is essential to ensure continued access to security updates and to remain compliant with modern platform standards.

Is there a timeline or plan for .NET 8 support?

### Proposed solution

Always support at least one LTS version

### Alternatives considered

There is currently no practical alternative that avoids the security implications of using an unsupported .NET version.

fbiel:
Hi,

could you share any updates or a timeline for .NET 8 compatibility? It's been 10 days without any reaction.

Again, this is a serious security liability, and we can’t move forward with our release on an unsupported platform.

Kwok-he-Chu:
I'll work on this as part of the next release in end of May. We already have opened an issue internally

Issue #1129, opened 2025-03-31
[Bug]: can't use Adyen .NET v30 / 31 on arm32v7


the-sutt:
### Description

When building a docker image for arm32v7, the Adyen .NET library crashes during requests.

### Steps to reproduce

1. Create an app that sends a diagnosis request
2. pack that app with docker and build with `.NET SDK 8`
3. execute on a 32bit arm-system

Sample project and more elaborate explanation:
https://github.com/the-sutt/AdyenTest/

Library works under arm64 just fine, even with a arm32v7 build. But the same arm32v7 build does not work on a arm32 system.

### Actual behavior

During sending the request, the following exception is thrown:

```
Creating device...
Creating client...
Creating local ApiService...
Performing diagnosis...
Constructing request...
Sending request...
Unhandled exception. System.Runtime.InteropServices.COMException (0x8007054F): An internal error occurred.
 (0x8007054F)
   at System.Threading.WaitHandle.WaitOneCore(IntPtr waitHandle, Int32 millisecondsTimeout)
   at System.Threading.WaitHandle.WaitOneNoCheck(Int32 millisecondsTimeout)
   at System.Threading.TimerQueue.TimerThread()
   at System.Threading.Thread.StartCallback()
Fatal error. Internal CLR error. (0x80131506)
   at System.Threading.WaitHandle.WaitOneCore(IntPtr, Int32)
   at System.Threading.WaitHandle.WaitOneNoCheck(Int32)
   at System.Threading.TimerQueue.TimerThread()
   at System.Threading.Thread.StartCallback()
```

### Expected behavior

Well... it should perform the request ;)

### Code snippet or screenshots (if applicable)

Please see sample project under https://github.com/the-sutt/AdyenTest/

### Adyen .NET API Library version

31

### .NET version

8

### Operating System

Linux

Kwok-he-Chu:
The library currently [only supports .NET 6](https://github.com/Adyen/adyen-dotnet-api-library/blob/v31.0.0/Adyen/Adyen.csproj#L4). It's most likely that one of the dependencies may need work well in a 32-bit environment. Could you try changing your [DockerFile](https://github.com/the-sutt/AdyenTest/blob/master/Dockerfile#L1) to build with the .NET 6.0 image instead?

Note: The plan to support .NET 8.0 is on the way

Kwok-he-Chu:
[...]

PR #1141, opened 2025-05-08
Upgrade to .NET 8.0 & added support for net462

Kwok-he-Chu:
This PR addresses the library upgrade to .NET 8.0

### Changes to HttpClient
* Developers can now set `ConnectTimeout`(default: 1 minute), `PooledConnectionLifetime`, `PooledConnectionIdleTimeout`, `MaxConnectionsPerServer`

* .NET library now uses SocketsHttpHandler in .NET6+ for better performance, but default to `HttpClientHandler` in older versions of .NET

### .NET 8.0 upgrade
* Added support for multiple .NET frameworks, on top of the regular `netstandard2.0` & `.NET 6.0`
  * `.NET 8.0` (LTS)
  * `net462`

* The language `LangVersion` of C# has been updated to `12` 

* `Nullable` property is set to **disable**d:
   * The `<Nullable>` feature has been explicitly disabled, ensuring backward compatibility and getting rid of a bunch of warnings until the [adyen-sdk-generator](https://github.com/Adyen/adyen-sdk-automation) gets upgraded.

* `<ImplicitUsings>` property is set to **enable**d
    * Due to reserved keywords such as `System.Environment`, some classes that require `Adyen.Environment` will have namespace conflicts due to global usings. I have addressed this by importing the namespaces manually

* Tagged the `Adyen.Test` & `Adyen.IntegrationTest` as a "TestProject" using .NET 6.0 & .NET 8.0

* Cleanups for unused properties & added some minor comments

### DIFF
--- a/.github/workflows/build-net6.0.yml
+++ b/.github/workflows/build-net6.0.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Setup .NET 6.0
         uses: actions/setup-dotnet@v4
         with:
-          dotnet-version: 6.0.x
+          dotnet-version: [ 6.0.x, 8.0.x ] # Adyen, Adyen.Test and Adyen.IntegrationTest targets both .NET 8.0 and .NET 6.0
         
       - name: Build (debug) in .NET 6.0
         run: dotnet build -c Debug -f net6.0
@@ -58,7 +58,7 @@ jobs:
       - name: Setup .NET 6.0
         uses: actions/setup-dotnet@v4
         with:
-          dotnet-version: 6.0.x
+          dotnet-version: [ 6.0.x, 8.0.x ] # Adyen, Adyen.Test and Adyen.IntegrationTest targets both .NET 8.0 and .NET 6.0
         
       - name: Build (debug) in .NET 6.0
         run: dotnet build -c Debug -f net6.0
--- a/.github/workflows/build-net6.0.yml
+++ b/.github/workflows/build-net6.0.yml
@@ -28,13 +28,13 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@v4
         with:
-          dotnet-version: 8.0
+          dotnet-version: 6.0.x
         
       - name: Build Debug
-        run: dotnet build -c Debug -f net8.0
+        run: dotnet build -c Debug -f net6.0
 
-      - name: Run unit tests on net8.0
-        run: dotnet test --no-build -c Debug -f net8.0 Adyen.Test/Adyen.Test.csproj
+      - name: Run unit tests on net6.0
+        run: dotnet test --no-build -c Debug -f net6.0 Adyen.Test/Adyen.Test.csproj
 
 
   integration-tests:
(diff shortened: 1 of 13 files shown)


Kwok-he-Chu:
Merging this closes: #1131 and #1129

.github/workflows/publish.yml
```diff
@@ -7,28 +7,46 @@ on:
...
+      
+      # Build in Release configuration and run unit tests in .NET 6.0
       - name: Build Release
         run: dotnet build Adyen -c Release
-      - name: Run test suite netcoreapp2.1
-        run: dotnet test --no-build -c Debug -f netcoreapp2.1 Adyen.Test/Adyen.Test.csproj
-      - name: Run test suite net6.0
-        run: dotnet test --no-build -c Debug -f net6.0 Adyen.Test/Adyen.Test.csproj
+
+      - name: Run unit tests on ne68.0
```
gcatanese, line 30: Minor error here `ne68.0`, should be `net6.0`?

.github/workflows/publish.yml
```diff
@@ -7,28 +7,46 @@ on:
...
+        run: dotnet test --no-build -c Release -f net6.0 Adyen.Test/Adyen.Test.csproj
+
+      # Set up .NET 8.0
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: 6.0.x
```
gcatanese, line 37: Should this be `NET 8`?

.github/workflows/build-net6.0.yml
```diff
@@ -21,28 +21,32 @@ jobs:
...
-        dotnet-version: [ '6.0.x' ]
 
     steps:
       - uses: actions/checkout@v4
       
       - name: Setup .NET
         uses: actions/setup-dotnet@v4
         with:
-          dotnet-version: ${{ matrix.dotnet-version }}
+          dotnet-version: 6.0.x
```
gcatanese, line 31: is it a good idea to use `6.0.x`? isn't better/safer to lock a specific version?

.github/workflows/publish.yml
```diff
@@ -7,28 +7,46 @@ on:
...
   # publish the nuget package
   publish:
-    name: build, pack & publish
-    runs-on: ubuntu-latest
+    name: Build, Pack & Publish
+    runs-on: ${{ matrix.os }}
+    
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, windows-latest ]
```
gcatanese, line 15: With multiple values in the matrix the workflow steps below are executed twice, it looks like that we publish twice (o one for each value in the matrix). 
Remove the matrix (maybe os compatibility can be checked in the build workflows) or create a different step (independent, no matrix) for the actual Nuget publish

gcatanese (changes requested):

.github/workflows/publish.yml
Kwok-he-Chu: Good catch

.github/workflows/build-net6.0.yml
Kwok-he-Chu: Let's keep the .NET version up-to-date with the latest security-patches: https://dotnet.microsoft.com/en-us/download/dotnet/8.0

gcatanese (approved):

PR #1147, opened 2025-05-14
Populate ApplicationName in UserAgent when not null

Kwok-he-Chu:
Fix: Populates ApplicationName in UserAgent when not null. Previously this value could be `null` if `ApplicationName` was not set in the `Config`.

### DIFF
--- a/Adyen/HttpClient/HttpClientWrapper.cs
+++ b/Adyen/HttpClient/HttpClientWrapper.cs
@@ -65,7 +65,16 @@ namespace Adyen.HttpClient
             httpRequestMessage.Headers.Add("ContentType", "application/json");
             httpRequestMessage.Headers.Add("Accept-Charset", "UTF-8");
             httpRequestMessage.Headers.Add("Cache-Control", "no-cache");
-            httpRequestMessage.Headers.Add("UserAgent", $"{_config.ApplicationName} {ClientConfig.UserAgentSuffix}{ClientConfig.LibVersion}");
+
+            if (!string.IsNullOrWhiteSpace(_config.ApplicationName))
+            {
+                httpRequestMessage.Headers.Add("UserAgent", $"{_config.ApplicationName} {ClientConfig.UserAgentSuffix}{ClientConfig.LibVersion}");
+            }
+            else
+            {
+                httpRequestMessage.Headers.Add("UserAgent", $"{ClientConfig.UserAgentSuffix}{ClientConfig.LibVersion}");
+            }
+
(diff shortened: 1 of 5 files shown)


Adyen/HttpClient/HttpClientWrapper.cs
```diff
@@ -65,7 +65,16 @@ public HttpRequestMessage GetHttpRequestMessage(string endpoint, string requestB
...
-            httpRequestMessage.Headers.Add("UserAgent", $"{_config.ApplicationName} {ClientConfig.UserAgentSuffix}{ClientConfig.LibVersion}");
+
+            if (!string.IsNullOrWhiteSpace(_config.ApplicationName))
+            {
+                httpRequestMessage.Headers.Add("UserAgent", $"{_config.ApplicationName} {ClientConfig.UserAgentSuffix}{ClientConfig.LibVersion}");
+            }
+            else
+            {
+                httpRequestMessage.Headers.Add("UserAgent", $"{ClientConfig.UserAgentSuffix}{ClientConfig.LibVersion}");
+            }
```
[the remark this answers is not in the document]
Kwok-he-Chu: No thanks, unreadable.

Adyen/HttpClient/HttpClientWrapper.cs
```diff
@@ -65,7 +65,16 @@ public HttpRequestMessage GetHttpRequestMessage(string endpoint, string requestB
             httpRequestMessage.Headers.Add("ContentType", "application/json");
             httpRequestMessage.Headers.Add("Accept-Charset", "UTF-8");
             httpRequestMessage.Headers.Add("Cache-Control", "no-cache");
-            httpRequestMessage.Headers.Add("UserAgent", $"{_config.ApplicationName} {ClientConfig.UserAgentSuffix}{ClientConfig.LibVersion}");
+
+            if (!string.IsNullOrWhiteSpace(_config.ApplicationName))
+            {
+                httpRequestMessage.Headers.Add("UserAgent", $"{_config.ApplicationName} {ClientConfig.UserAgentSuffix}{ClientConfig.LibVersion}");
```
gcatanese, line 71: The `ClientConfig.UserAgentSuffix` can have a better name (LibName?) and it shouldn't include `/`, I think this didn't get updated in this PR, but maybe it can be addressed.

Kwok-he-Chu: The `/` is here: https://github.com/Adyen/adyen-dotnet-api-library/blob/v32.0.0/Adyen/Constants/ClientConfig.cs#L14 and i don't want to change this value as it would be a `breaking-change`

PR #1170, opened 2025-07-16
[Checkout] Skip serialisation of attributes with default values

gcatanese:
## Description
This PR prevents attributes with default values from being included in the serialised API request payload unless explicitly set. This avoids sending redundant fields that already handled by the API backend (matching the defaults defined in the OpenAPI spec).

## Changes
Updated the `Mustache` template to skip initialising constructor parameters with default values.

From
```
bool? threeDSAuthenticationOnly = false
```
to
```
bool? threeDSAuthenticationOnly = null
```

**Note**: if the constructor parameter is required then then it is initialised with the default value (if provided)

Regenerated Checkout models to reflect the change (other models will be handled by the SDK Automation Bot).

Added unit tests to verify the behaviour.

**Note**: existing tests in `PaymentMethodDetailsTest` have been updated to set explicitly the payment method `type`. Although not required in the OpenAPI spec, applications are expected to set the value when performing a payment.


Adyen/Model/Checkout/RakutenPayDetails.cs
```diff
@@ -61,7 +61,7 @@ public enum TypeEnum
         /// <param name="recurringDetailReference">This is the &#x60;recurringDetailReference&#x60; returned in the response when you created the token..</param>
         /// <param name="storedPaymentMethodId">This is the &#x60;recurringDetailReference&#x60; returned in the response when you created the token..</param>
         /// <param name="type">**rakutenpay** (default to TypeEnum.Rakutenpay).</param>
```
Kwok-he-Chu, line 63: The default is now set to `null`, can we find a way to initialize `TypeEnum.Rakutenpay` in the constructor or as a separate field?

Adyen.Test/PaymentMethodDetailsTest.cs
```diff
@@ -11,6 +11,7 @@ public void TestAchPaymentMethod()
         {
             var achDetails = new AchDetails
             {
+                Type = AchDetails.TypeEnum.Ach,
```
Kwok-he-Chu, line 14: Marked as a **breaking change** - Developers will have to add this property after this change. 
Suggestion: Can we make auto-populate this property? It seems like it could only have one possible value.

Adyen/Adyen.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   
   <PropertyGroup>
+    <RootNamespace>Adyen</RootNamespace>
```
Kwok-he-Chu, line 4: Not needed, it would take the `Adyen`-.csproj as default

Adyen/Model/Checkout/RakutenPayDetails.cs
Kwok-he-Chu: Can we confirm that the backend defaults to the first enum value (e.g. Rakutenpay)?

PR #1177, opened 2025-08-07
Mustache template with unescaped HTML

gcatanese:
`<summary>` and `<value>` are not correctly displayed, special characters are HTML escaped.

This PR updates the templates to render the content of the description without escaping, using `{{{description}}}`


gcatanese:
To be considered as part of the OpenAPI Generator upgrade

PR #1235, opened 2026-01-26
feat: handle unknown enum values gracefully

galesky-a:
# Description
This PR introduces a "graceful fail" mechanism for enum deserialization. It ensures that when the Adyen API adds new enum values, the SDK remains functional by treating unknown values as null rather than throwing a JsonSerializationException.

## Related PRs and runs:

- [Generated code](https://github.com/Adyen/adyen-dotnet-api-library/pull/1248) triggered by this [action](https://github.com/Adyen/adyen-sdk-automation/actions/runs/21757352562)
- [SDK Automation](https://github.com/Adyen/adyen-sdk-automation/pull/97) to target this branch and skip unsupported fields (ex: `oneOf`, `allOf`)
- Tests run for [Dotnet 8.0](https://github.com/Adyen/adyen-dotnet-api-library/actions/runs/21757540003) and [6.0](https://github.com/Adyen/adyen-dotnet-api-library/actions/runs/21757547484)

## Key Changes
- Added SafeStringEnumConverter, which extends StringEnumConverter. It catches serialization exceptions for unknown enum members and returns null for nullable enum types.
- Registered the new converter in JsonOperation.Deserialize<T>() to apply it across all SDK operations.
- Modified modelGeneric.mustache to ensure all generated enum properties and constructor arguments are nullable, **regardless of whether they are "required" in the API specification.**. This is the most impactful change, this should always have been the case, it's the behavior the API was modeled to have, hence this change is considered a bugfix

## Testing Infrastructure:

- Added test suite (SafeStringEnumConverterTest.cs) to verify that unknown enum values deserialize to null while known values continue to work as expected.

## How to verify
- Run the new test suite: Adyen.Test/SafeStringEnumConverterTest.cs.
- Observe that providing an unknown string to a nullable enum field no longer crashes the JsonConvert process.


Adyen/Adyen.csproj
```diff
@@ -35,6 +35,7 @@
   </PropertyGroup>
   
   <ItemGroup>
+    <PackageReference Include="JsonSubTypes" Version="2.0.1" />
```
galesky-a: oops, removing this

Adyen.Test/UnknownEnumDeserializationTest.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            var json = JsonConvert.SerializeObject(paymentRequest);
+            
+            // Null values should not be serialized (based on EmitDefaultValue = false)
+            Assert.IsFalse(json.Contains("\"channel\""));
+            Assert.IsFalse(json.Contains("\"entityType\""));
+        }
+
+        [TestMethod]
+        public void TestWebhookHandlerWithUnknownEnumValues()
```
gcatanese: Does this test need to be renamed? It seems it doesnt test enums

gcatanese (approved):
Left one minor comment

Adyen.Test/UnknownEnumDeserializationTest.cs
galesky-a: Renamed!

Adyen.Test/NewtonsoftEnumBehaviorTest.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+            catch (Exception ex)
+            {
+                exception = ex;
+            }
+            
+            if (exception == null)
+            {
+                var result = JsonConvert.DeserializeObject<TestModel>(json);
+                Assert.IsNull(result.EnumField);
```
Kwok-he-Chu: Line 47-48 ` Assert.IsNull(result.EnumField);` is never reached, because the function will always thrown an exception.

Suggestion: Remove these lines (45-50) where `exception == null`

**Rename**: Given_`StringEnumConverter`_When_`UnknownValue`_Throws_`JsonSerializationException`
and **create** a new test to assert that lines 45-50 separately:

```csharp
[TestMethod]
  public void Given_StringEnumConverter_When_Null_Result_TestModel_EnumField_Is_Null()
  {
      var json = @"{""enumField"": null}";
      
      var result = JsonConvert.DeserializeObject<TestModel>(json);
      Assert.IsNull(result.EnumField);
  }
```

Adyen.Test/NewtonsoftEnumBehaviorTest.cs
```diff
@@ -0,0 +1,67 @@
+using System;
```
Kwok-he-Chu: nit: rename filename to `NewtonsoftEnumBehaviorTest` to `StringEnumConverterTest`, This tests tests how the `StringEnumConverter` behaves

Adyen.Test/NewtonsoftEnumBehaviorTest.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.Runtime.Serialization;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Adyen.Test
+{
+    [TestClass]
+    public class NewtonsoftEnumBehaviorTest
+    {
+        [JsonConverter(typeof(StringEnumConverter))]
+        public enum TestEnum
```
Kwok-he-Chu: nit: scope - Make TestEnum private so it's only accessible from `Adyen.Test.NewotnsoftEnumBehaviorTest`

Adyen.Test/NewtonsoftEnumBehaviorTest.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            [EnumMember(Value = "Value1")]
+            Value1 = 1,
+            
+            [EnumMember(Value = "Value2")]
+            Value2 = 2
+        }
+
+        public class TestModel
```
Kwok-he-Chu: nit: scope - Make TestModel private so it's only accessible from `Adyen.Test.NewotnsoftEnumBehaviorTest`

Adyen.Test/NewtonsoftEnumBehaviorTest.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+            else
+            {
+                // If this throws, we need a custom converter
+                Assert.IsInstanceOfType(exception, typeof(JsonSerializationException));
+            }
+        }
+
+        [TestMethod]
+        public void TestNewtonsoft_KnownEnumValue_WithNullable()
```
Kwok-he-Chu: Rename using the `Given`_1_`When`_2_Then`_3_-approach

Suggestion: Given_`StringEnumConverter`_When_`Value1`_Returns_`Value1_As_Enum`

Kwok-he-Chu: nit: Name method to reflect what it's asserting, prefer: using the `Given`_1_`When`_2_`Then`_3_-approach 

Suggestion: Given_`StringEnumConverter`_When_`Value1`_Returns_`Value1_As_Enum`

Adyen/Util/SafeStringEnumConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+                }
+                
+                // For non-nullable enums, we still throw to maintain backward compatibility
+                throw;
+            }
+        }
+
+        private static bool IsNullableType(Type type)
+        {
+            return type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Nullable<>);
```
Kwok-he-Chu: I'm not sure what the StringEnumConverter passes as the underlying enum value, so if `Nullable<T>` works, go ahead- I'd suggest replacing it with the following (also used in Newtonsoft.Json itself):

https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Converters/StringEnumConverter.cs#L228
```csharp
Nullable.GetUnderlyingType(type) != null; 
```

Adyen/Util/JsonOperation.cs
```diff
@@ -14,6 +14,7 @@ public static T Deserialize<T>(string response)
         {
             var jsonSettings = new JsonSerializerSettings();
             jsonSettings.Converters.Add(new ByteArrayConverter());
+            jsonSettings.Converters.Add(new SafeStringEnumConverter());
```
Kwok-he-Chu: This line-17 doesn't do anything if you add the `SafeStringEnumConverter` globally, e.g.:

```csharp

[JsonConverter(typeof(Adyen.Util.SafeStringEnumConverter))]
///... your enum class
```

* [ ] Can you verify if you remove line-17 that the tests should still **all** pass?

Adyen/Util/JsonOperation.cs
```diff
@@ -14,6 +14,7 @@ public static T Deserialize<T>(string response)
...
```
Kwok-he-Chu, line 17: `[JsonConverter(typeof(Adyen.Util.SafeStringEnumConverter))]`

Missing: I'm trying to find the change above in the `xxx.mustache` template (this adds the JsonConverter globally), can you push the mustache file for completeness?

Adyen/Util/SafeStringEnumConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    public class SafeStringEnumConverter : StringEnumConverter
+    {
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            try
+            {
+                return base.ReadJson(reader, objectType, existingValue, serializer);
+            }
+            catch (JsonSerializationException)
```
Kwok-he-Chu, line 19: Incomplete: cover the generic Exception, this will otherwise not handle other exceptions

```csharp

 catch (Exception exception) when (
            exception is JsonSerializationException ||
            exception is ...
        )
```

Adyen/Util/JsonOperation.cs
galesky-a: I'm changing it in post-processing actually:
https://github.com/Adyen/adyen-sdk-automation/pull/97/changes#diff-f020f28f2856d80e470a637439d130165ef76f13db9ca491858c367e30bdf630R42

Reason being that the default `StringEnumConverter` is also not set on the mustache template explicitly.

Adyen/Util/SafeStringEnumConverter.cs
galesky-a: Should it though ?

It's usually best to narrow what you catch, otherwise it can hide bugs (like NullReferenceException or StackOverflowException) that should still probably bubble the exception.

Adyen/Util/JsonOperation.cs
galesky-a: Verified, removed line 17. All 311 tests pass. The `[JsonConverter(typeof(SafeStringEnumConverter))]` attribute on each enum definition already handles this. Worth noting this was a safety net for any enum that might miss the attribute, but in hindsight since these are code-generated, the risk is minimal.

Adyen/Util/SafeStringEnumConverter.cs
galesky-a: Huh, nice one!

I think they are functionally equivalent in this case but makes sense to use something more idiomatic + same as Newton.

Adyen/Util/SafeStringEnumConverter.cs
Kwok-he-Chu: Can you check which other exceptions `ReadJson` can throw? (leave out: NullReferenceException or StackOverflowException etc.)

Adyen/Util/JsonOperation.cs
Kwok-he-Chu: That makes sense why I wasn't able to find it, thanks!

Adyen/Util/SafeStringEnumConverter.cs
galesky-a: Just checked, `ReadJson`  can only throw `JsonSerializationException`. Every throw path produces that same type. It even has a `catch (Exception ex)` inside.

galesky-a:
> This is not an easy change to make, great job. Added feedback (nits, missing-file **?** & suggestions) based on these changes.
> 
> **Note:** With this nullable change, developers have to check whether the EnumField is null (or check `EnumField.HasValue`) before accessing the underlying enum value.
> 
> * [ ]  Have you also confirmed that the `ToString()`-concatenation in the models would behave when the enum is null? @galesky-a


Yes, confirmed. `StringBuilder.Append()` with a null nullable enum simply appends nothing (empty string). `Nullable.Equals()` and `Nullable.GetHashCode()` are also safe since `Nullable<T>` acts as struct, `.Equals()` returns false (or true if both null), and `.GetHashCode()` returns 0 when HasValue is false. No NullReferenceException risk in any of the generated model methods (ToString(), Equals(), GetHashCode()).

That said, the broader point is valid: consumers of the library will need to check HasValue/null before using the enum value in switch statements, comparisons, etc. This is an expected trade-off for forward compatibility with unknown enum values.

PR #1296, opened 2026-02-10
Recurring: Correct BASE_URL for Adyen API endpoint

gcatanese:
Correct Recurring base url (value in the spec is incorrect)

### DIFF
--- a/Adyen/Recurring/Client/HostConfiguration.cs
+++ b/Adyen/Recurring/Client/HostConfiguration.cs
@@ -40,7 +40,7 @@ namespace Adyen.Recurring.Client
         /// The base path of the API, it includes the http(s)-scheme, the host domain name, and the base path.
         /// This value can change when `ConfigureAdyenOptions` is called in <see cref="HostConfiguration"/>). The new value will be based on the <see cref="AdyenOptions"/>.<see cref="AdyenEnvironment"/>.
         /// </summary>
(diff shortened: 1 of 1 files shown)


Adyen/Recurring/Client/HostConfiguration.cs
```diff
@@ -40,7 +40,7 @@ public class HostConfiguration
         /// The base path of the API, it includes the http(s)-scheme, the host domain name, and the base path.
         /// This value can change when `ConfigureAdyenOptions` is called in <see cref="HostConfiguration"/>). The new value will be based on the <see cref="AdyenOptions"/>.<see cref="AdyenEnvironment"/>.
         /// </summary>
-        public static string BASE_URL = "https://paltokenization-test.adyen.com/pal/servlet/Recurring/v68";
+        public static string BASE_URL = "https://pal-test.adyen.com/pal/servlet/Recurring/v68";
```
Kwok-he-Chu, line 43: This BASE_URL is no longer used: https://github.com/Adyen/adyen-dotnet-api-library/blob/main/Adyen/Checkout/Services/PaymentLinksService.cs#L320 - It's now a hardcoded value on service-level @gcatanese 

BASE_URL can be removed

gcatanese: This constant is no longer in the code

PR #1333, opened 2026-02-13
[Patch 32.2.2] Code generation: update services and models

AdyenAutomationBot:
OpenAPI spec files or templates have been modified on 13-02-2026         by [commit](https://github.com/Adyen/adyen-openapi/commit/a9d350f1449f3fbce28ab67f6eb6a4b9904be640).


Adyen/Model/Checkout/UpiQrDetails.cs
```diff
@@ -1,247 +0,0 @@
-/*
```
galesky-a, line 1: @gcatanese I remember some discussion on UPI, can you double check this is intended?

galesky-a: https://github.com/Adyen/adyen-dotnet-api-library/pull/1198/changes#diff-1ddc14718a9ea9872ba42039de3274dcf5507be4e5672b1e69d8eb6b6dcea9db

PR #1362, opened 2026-02-16
Verify Recurring test and live URLs

gcatanese:
Improve service generation, extracting the `BASE_URL` in a constant

Add test to verify `test` and `live` URLs are correct

### DIFF
--- a/Adyen.Test/Recurring/RecurringHttpClientBaseAddressTest.cs
+++ b/Adyen.Test/Recurring/RecurringHttpClientBaseAddressTest.cs
@@ -13,11 +13,12 @@ namespace Adyen.Test.Recurring
     [TestClass]
     public class RecurringHttpClientBaseAddressTest
     {
-        private readonly JsonSerializerOptionsProvider _jsonSerializerOptionsProvider;
-        private readonly IRecurringService _recurringService;
-
-        public RecurringHttpClientBaseAddressTest()
-        { 
+        
+        
+        [TestMethod]
+        public async Task RecurringService_TestUrl_IsCorrect()
+        {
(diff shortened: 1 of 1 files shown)


Adyen.Test/Recurring/RecurringHttpClientBaseAddressTest.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    [TestClass]
+    public class RecurringHttpClientBaseAddressTest
+    {
+        private readonly JsonSerializerOptionsProvider _jsonSerializerOptionsProvider;
+        private readonly IRecurringService _recurringService;
+
+        public RecurringHttpClientBaseAddressTest()
+        { 
+            IHost host = Host.CreateDefaultBuilder()
```
Kwok-he-Chu, line 22: nit: Move IHost from member-scope to (line 21-28) to local-scope (line 39)

Adyen/Recurring/Models/Amount.cs
```diff
@@ -80,28 +80,6 @@ public override string ToString()
             sb.Append("}\n");
             return sb.ToString();
         }
-
-        /// <summary>
-        /// To validate all properties of the instance
```
galesky-a, line 85: Are those changes to validation intentional too ?

gcatanese: Validation was removed as part of #1237, but the code was not (re)generated for Recurring. 
In this PR I have re-generated the models to create the tests, knowing it is a temporary generation, we are updating the Mustache templates that will re-trigger another full generation.

PR #1475, opened 2026-04-09
Fix: Json constructor stj deserialization

### LINKED ISSUE #1474: [Bug]: `PaymentMethodsResponse` deserialization fails in Checkout after the `System.Text.Json` migration introduced in `v34.0.0`
### Description

After upgrading to Adyen .NET API Library `v34.x`, we are seeing a deserialization regression in a Checkout model after the library-wide `System.Text.Json` migration introduced in `v34.0.0`.  Our concrete reproducible case is `PaymentMethodsAsync`: Adyen returns HTTP `200 OK`, but the Adyen .NET API Library fails while deserializing the successful response into `Adyen.Checkout.Models.PaymentMethodsResponse`.  The failure appears to be caused by the `PaymentMethodsResponse` model shape in `34.0.2`. The type has a `[JsonConstructor]` whose parameters use `Option<List<...>>`, while the JSON-facing properties are `List<...>`. `System.Text.Json` rejects this constructor binding.  This may be related to the broader STJ migration already visible elsewhere in the library, but this report is specifically about a Checkout model deserialization failure in `34.0.2`.

### Steps to reproduce

1. Install Adyen .NET API Library `34.0.2`.
2. Call Checkout `paymentMethods` through `PaymentsService.PaymentMethodsAsync(...)`.
3. Receive a valid HTTP `200 OK` response from Adyen.
4. Call `TryDeserializeOkResponse(out PaymentMethodsResponse result)`.
5. Observe deserialization failure.

Minimal reproduction without making a real HTTP request:

1. Install Adyen .NET API Library `34.0.2`.
2. Deserialize the following JSON into `Adyen.Checkout.Models.PaymentMethodsResponse` using `System.Text.Json`.
3. Observe `InvalidOperationException`.

### Actual behavior

Deserialization fails with:

```text
System.InvalidOperationException: Each parameter in the deserialization constructor on type 'Adyen.Checkout.Models.PaymentMethodsResponse' must bind to an object property or field on deserialization. Each parameter name must match with a property or field on the object. Fields are only considered when 'JsonSerializerOptions.IncludeFields' is enabled. The match can be case-insensitive.
```

In the Adyen SDK logging this appears as:

```text
An error occurred while deserializing the "OK" response.
```

### Expected behavior

If Adyen returns HTTP `200 OK` for `/paymentMethods`, the library should deserialize the response into `PaymentMethodsResponse` successfully.

More generally, the `System.Text.Json` migration introduced in `v34.0.0` should not leave Checkout response models in a state where they cannot be deserialized by the library itself.

### Code snippet or screenshots (if applicable)

Runtime path:

```csharp
var apiResponse = await checkout.PaymentMethodsAsync(paymentMethodsRequest, null, default);

if (!apiResponse.TryDeserializeOkResponse(out PaymentMethodsResponse response))
{
    throw new Exception("Failed to deserialize Adyen payment methods response.");
}
```

Minimal reproduction:

```csharp
using System.Text.Json;
using Adyen.Checkout.Models;

var json = "{\"paymentMethods\":[],\"storedPaymentMethods\":[]}";

var result = JsonSerializer.Deserialize<PaymentMethodsResponse>(json);
```

Relevant model shape observed from `34.0.2`:

```text
[JsonConstructor]
PaymentMethodsResponse(
    Option<List<PaymentMethod>> paymentMethods,
    Option<List<StoredPaymentMethod>> storedPaymentMethods)

public List<PaymentMethod> PaymentMethods { get; set; }            // json name: paymentMethods
public List<StoredPaymentMethod> StoredPaymentMethods { get; set; } // json name: storedPaymentMethods
```

This seems incompatible with `System.Text.Json` constructor binding.

### Adyen .NET API Library version

34.0.2

### .NET version

.NET 8

### Operating System

Windows

### Additional context

- The HTTP request itself succeeds and Adyen returns `200 OK`.
- This appears to be a model/deserialization issue, not a request formatting issue.
- We reproduced the problem locally without calling the real API by deserializing a minimal JSON payload directly into `PaymentMethodsResponse`.
- `PropertyNameCaseInsensitive = true` does not fix it.
- `IncludeFields = true` does not fix it.
- Forcing `System.Text.Json` to use the parameterless constructor for this type does work, which suggests the problem is specifically the `[JsonConstructor]` path on `PaymentMethodsResponse`.
- `v34.0.0` release notes explicitly mention:
  `Support native System.Text.Json for serialization and deserialization.`
- We do not see anything in the `v34.0.0` release notes that suggests consumers should apply a special Checkout-specific migration step for `PaymentMethodsResponse`.
- `34.0.2` release notes do not mention a fix related to Checkout model deserialization; they only mention:
  `Fix Classic Recurring live URL (#1444)`

.Net Fiddle: https://dotnetfiddle.net/eaXF1g

jorismathijssen:
## What problem this solves

After upgrading to v34.x, Checkout operations such as `PaymentMethodsAsync` return HTTP 200 OK
but the SDK immediately fails to deserialize the response, throwing:

```
System.InvalidOperationException: Each parameter in the deserialization constructor on type
'Adyen.Checkout.Models.PaymentMethodsResponse' must bind to an object property or field on
deserialization.
```

This surfaces as `"An error occurred while deserializing the OK response"` in SDK logging and
causes `TryDeserializeOkResponse` to return `false` even for a valid API response.

Minimal reproduction without a real HTTP call:

```csharp
// Throws InvalidOperationException in v34.0.x
var result = JsonSerializer.Deserialize<PaymentMethodsResponse>(
    "{\"paymentMethods\":[],\"storedPaymentMethods\":[]}");
```

## Root cause

The `modelGeneric.mustache` template placed `[JsonConstructor]` on the primary constructor of
every generated model. For models with optional fields the constructor takes `Option<T>` parameters:

```csharp
// BROKEN
[JsonConstructor]
public PaymentMethodsResponse(
    Option<List<PaymentMethod>?> paymentMethods = default,
    Option<List<StoredPaymentMethod>?> storedPaymentMethods = default)
```

The JSON-facing properties expose plain `T`:

```csharp
[JsonPropertyName("paymentMethods")]
public List<PaymentMethod>? PaymentMethods { get; set; }
```

System.Text.Json validates at type-metadata resolution time that every `[JsonConstructor]`
parameter can be matched **by name and type** to a public property. `Option<List<PaymentMethod>?>`
≠ `List<PaymentMethod>?`, so STJ throws `InvalidOperationException`.

Every model already ships a dedicated `JsonConverter` (registered via `HostConfiguration`) whose
`Read` method calls the `Option<T>` constructor explicitly — `[JsonConstructor]` was never needed
for correct behaviour, only harmful.

## My approach

Remove `[JsonConstructor]` from the template and from all 1,315 generated model files. When
STJ encounters a model without a registered converter (bare `JsonSerializer.Deserialize<T>` call),
it now falls back to the parameterless constructor + property setters, which works correctly.
The custom converters registered via `HostConfiguration` are unaffected and continue to take
full priority in the normal SDK path.

## Changes

| File | Change |
|---|---|
| `templates-v7/csharp/libraries/generichost/modelGeneric.mustache` | Remove `[JsonConstructor]` emission block — prevents regression on every future codegen run |
| `Adyen/**/Models/*.cs` (1,315 files) | Remove `[JsonConstructor]` attribute from generated constructors |
| `Adyen.Test/Checkout/PaymentsTest.cs` | Add 3 regression tests |

## Backward compatibility

**No breaking changes.** Verified across every caller pattern and JSON shape:

| Caller pattern | v34.0.x | After fix |
|---|---|---|
| `ConfigureCheckoutDefaults` + `TryDeserializeOkResponse` | **Broken** | ✅ Fixed — custom `JsonConverter` path unchanged |
| `JsonSerializer.Deserialize<T>(json, sdkOptions)` with HostConfiguration options | **Broken** (some STJ contexts) | ✅ Fixed — converter still takes priority |
| `JsonSerializer.Deserialize<T>(json)` bare, no options | **Broken** — always threw | ✅ Fixed — uses parameterless constructor + property setters |
| Direct construction `new PaymentMethodsResponse(...)` | ✅ Worked | ✅ Unchanged — constructor itself is not touched |
| Serialization `JsonSerializer.Serialize(...)` | ✅ Worked | ✅ Unchanged — `[JsonConstructor]` only affects deserialization |
| `Option<T>.IsSet` tracking | ✅ Worked via converter | ✅ Correct in both paths — property setter calls `new Option<T>(value)` |

**Old and new JSON shapes both work:**
- Missing optional fields → setter never called → `Option<T>.IsSet` remains `false` ✅
- Present optional fields → setter called → `Option<T>.IsSet` set to `true` ✅
- Empty arrays (`[]`) → setter called with empty list ✅
- Required fields → `{ get; set; }` on all required properties — STJ can set them via property setters ✅
- Read-only properties (3 found across all models) → all marked `[JsonIgnore]` or API-response-only — never deserialized from JSON ✅

**Why the parameterless-constructor fallback is safe:**
The template always generates a parameterless constructor when a model has any fields.
Required fields use `{ get; set; }`, optional fields use an `Option<T>` setter — both are
settable by STJ through normal property assignment after construction.

## Tests

Three regression tests added to `Adyen.Test/Checkout/PaymentsTest.cs`:

- **`Given_PaymentMethodsResponse_When_Deserialized_With_Default_Options_Then_Does_Not_Throw`**
  Exact minimal repro from #1474 — bare `JsonSerializer.Deserialize<PaymentMethodsResponse>`
  no longer throws.

- **`Given_PaymentMethodsResponse_With_StoredPaymentMethods_When_Deserialized_Then_All_Properties_Are_Populated`**
  Both `paymentMethods` and `storedPaymentMethods` are populated using the SDK's
  `JsonSerializerOptions` (HostConfiguration converter path).

- **`Given_PaymentMethod_When_Deserialized_With_Default_Options_Then_Does_Not_Throw`**
  `PaymentMethod` — the other model called out in the issue — is equally fixed.

Full test run: **412 passed, 0 failed**.

Fixes #1474

Disclaimer: AI was used to assist in drafting the documentation and PR description. The code, analysis, and fix itself were written and validated manually.


gcatanese:
@jorismathijssen Thank you very much for reporting the issue, explaining clearly the impact and providing us with a comprehensive fix, this is awesome work!

Unfortunately we cannot merge the PR as-is because we want to generate the code using our code generator (built on top of the OpenAPI Generator), however I want to check with you if you are still keen on authoring those changes (we love to have new contributors, and you deserve here the recognition).

If this is the case, the next steps are:
- in this PR only update the Mustache templates
- once the above is merged (*), the SDK Automation Bot will regenerate all models, following the changes in the Mustache templates
- create a new PR with the new tests 

(*)  we are processing several new issues related to `v34`, it looks like we might need to release a patch. If that's the case, we would hold off a little longer before merging this, but it shouldn't be very long).

Thank you again for the awesome contribution!

gcatanese:
Superseded by #1513 and #1517

PR #1517, opened 2026-05-01
Add regression tests for #1474 ([JsonConstructor] STJ deserialization)

### LINKED ISSUE #1474: [Bug]: `PaymentMethodsResponse` deserialization fails in Checkout after the `System.Text.Json` migration introduced in `v34.0.0`

gcatanese:
## Description

Adds 7 regression tests to `Adyen.Test/Checkout/PaymentsTest.cs` to cover the `[JsonConstructor]` + STJ deserialization bug reported in #1474.

### Tests from PR #1475

- `Given_PaymentMethodsResponse_When_Deserialized_With_Default_Options_Then_Does_Not_Throw` — exact minimal repro: bare `JsonSerializer.Deserialize<PaymentMethodsResponse>` no longer throws `InvalidOperationException`
- `Given_PaymentMethodsResponse_With_StoredPaymentMethods_When_Deserialized_Then_All_Properties_Are_Populated` — verifies both `paymentMethods` and `storedPaymentMethods` are populated using the SDK's `JsonSerializerOptions`
- `Given_PaymentMethod_When_Deserialized_With_Default_Options_Then_Does_Not_Throw` — proves the fix covers `PaymentMethod` (not just `PaymentMethodsResponse`)

### Additional tests

- `Given_PaymentMethodsResponse_When_OptionalFieldAbsent_With_Default_Options_Then_Field_Is_Omitted_On_Reserialize` — absent optional field stays absent when round-tripped through serialization, verifying `Option<T>.IsSet` remains `false` via the property-setter path
- `Given_PaymentMethodsResponse_When_OptionalFieldPresent_With_Default_Options_Then_Field_Is_Preserved_On_Reserialize` — present optional field is preserved on round-trip, verifying `Option<T>.IsSet` is set to `true` by the property setter
- `Given_PaymentMethodsResponse_When_OptionalFieldIsExplicitNull_With_Default_Options_Then_Does_Not_Throw` — JSON `null` for an optional field does not throw and yields a `null` property value
- `Given_ServiceError_When_Deserialized_With_Default_Options_Then_Does_Not_Throw` — extends regression coverage to a non-Checkout namespace (`AcsWebhooks.Models.ServiceError`), confirming the fix is not narrowly scoped to Checkout

## Fixed issue

Fixes #1474

### DIFF
--- a/Adyen.Test/AcsWebhooks/ServiceErrorTest.cs
+++ b/Adyen.Test/AcsWebhooks/ServiceErrorTest.cs
@@ -1,0 +1,32 @@
+using Adyen.AcsWebhooks.Models;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Text.Json;
+
+namespace Adyen.Test.AcsWebhooks
+{
+    [TestClass]
+    public class ServiceErrorTest
+    {
+        /// <summary>
+        /// Regression test for GitHub issue #1474 applied to a non-Checkout namespace.
+        /// Verifies that <see cref="ServiceError"/>, which has all-optional
+        /// fields, can be deserialized using default <see cref="JsonSerializerOptions"/> without throwing.
+        /// The <c>[JsonConstructor]</c> regression affected every generated model across all namespaces.
+        /// </summary>
+        [TestMethod]
+        public void Given_ServiceError_When_Deserialized_With_Default_Options_Then_Does_Not_Throw()
+        {
+            // Arrange
+            string json = "{\"errorCode\":\"000\",\"message\":\"test error\",\"status\":422}";
+
+            // Act
+            ServiceError result = JsonSerializer.Deserialize<ServiceError>(json);
+
+            // Assert
+            Assert.IsNotNull(result);
+            Assert.AreEqual("000", result.ErrorCode);
+            Assert.AreEqual("test error", result.Message);
+            Assert.AreEqual(422, result.Status);
+        }
+    }
+}
--- a/Adyen.Test/Checkout/PaymentsTest.cs
+++ b/Adyen.Test/Checkout/PaymentsTest.cs
@@ -950,6 +950,156 @@ namespace Adyen.Test.Checkout
                 "action must not be present when not set");
         }
 
+        /// <summary>
+        /// Regression test for GitHub issue #1474.
+        /// Verifies that <see cref="PaymentMethodsResponse"/> can be deserialized using default
+        /// <see cref="JsonSerializerOptions"/> without throwing <see cref="InvalidOperationException"/>.
+        /// The root cause was a <c>[JsonConstructor]</c> attribute on a constructor whose parameters
+        /// used <c>Option&lt;T&gt;</c> types that STJ could not bind to the plain <c>T</c> properties.
+        /// </summary>
+        [TestMethod]
+        public void Given_PaymentMethodsResponse_When_Deserialized_With_Default_Options_Then_Does_Not_Throw()
+        {
+            // Arrange - minimal repro from issue #1474
+            string json = "{\"paymentMethods\":[],\"storedPaymentMethods\":[]}";
+
+            // Act - previously threw InvalidOperationException about [JsonConstructor] parameter binding
+            PaymentMethodsResponse result = JsonSerializer.Deserialize<PaymentMethodsResponse>(json);
+
+            // Assert
+            Assert.IsNotNull(result);
+            Assert.IsNotNull(result.PaymentMethods);
+            Assert.IsNotNull(result.StoredPaymentMethods);
+            Assert.AreEqual(0, result.PaymentMethods.Count);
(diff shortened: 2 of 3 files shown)


gcatanese:
Addressed in 05a60757: moved `Given_ServiceError_When_Deserialized_With_Default_Options_Then_Does_Not_Throw` to `Adyen.Test/AcsWebhooks/ServiceErrorTest.cs` and also resolved the pre-existing merge conflict in `Adyen/AcsWebhooks/Models/ServiceError.cs`.

jeandersonbc (approved):

PR #1543, opened 2026-05-04
Remove Vagrantfile and Makefile

gcatanese:
## Description
Removes two files that are no longer relevant to this repository:

- **Makefile**: Was used to drive local OpenAPI code generation via a Java-based `openapi-generator-cli`. Code generation is now handled by the `adyen-sdk-automation` pipeline, so this file serves no purpose.
- **Vagrantfile**: Provided a local Ubuntu VM (via Parallels) to support the Makefile-based generation workflow. With the Makefile gone, this is equally obsolete.

## Tested scenarios
No functional code changes — purely removing stale tooling files.

## Fixed issue
N/A


jeandersonbc (approved):
Nice! I think we still have Vagrant files around some other repos that should be removed too.

PR #1680, opened 2026-05-12
Replace standalone C# enums with forward-compatible IEnum class pattern

### LINKED ISSUE #1612: [Bug] NullReferenceException on .NET 8 when deserializing unknown enum values in generated JsonConverter
## Context

When deserializing a JSON payload that contains an unknown/unrecognised enum value (e.g. `"type": "unknown-type"`), the generated `JsonConverter` throws a `NullReferenceException` on .NET 8.

The root cause is in the Mustache template (`templates-v7/csharp/libraries/generichost/JsonConverter.mustache`). For required enum properties it generates:

```csharp
balanceTransferResponse.Status = status.Value!;
```

`status.Value` is `StatusEnum?` and resolves to `null` when `FromStringOrDefault` returns null for an unknown value. On .NET 8, assigning this null via the null-forgiving operator to a non-nullable property causes a `NullReferenceException`. The same code works on .NET 9.

Affected generated files (as of this writing): `Adyen/BalanceControl/Models/BalanceTransferResponse.cs` and `Adyen/BalanceControl/Models/BalanceTransferRequest.cs`.

## Proposed Change

Update the Mustache template to emit a null-conditional assignment for required enum properties:

```csharp
// Instead of:
balanceTransferResponse.Status = status.Value!;

// Generate:
if (status.Value != null)
    balanceTransferResponse.Status = status.Value;
```

This ensures unknown enum values are silently treated as `null` rather than crashing.

The template fix is in the `{{#required}}` / `{{#isEnum}}` block of `JsonConverter.mustache`, and the corresponding generated files need to be regenerated.

## Benefit

Prevents a crash when the API returns an enum value not yet known to the SDK, improving forward-compatibility and stability on .NET 8.

### LINKED ISSUE #1650: Standalone C# enums are not forward-compatible with unknown API values
### Problem
Standalone C# enums (those defined at the top level of a model file, rather than nested) are not forward-compatible. When the Adyen API returns a new value that is not yet present in the SDK's enum definition, the JSON deserialization fails.

This manifests in two ways:
1. **Direct deserialization** of the enum type throws a `JsonException`.
2. **Model deserialization** containing the enum throws an `InvalidOperationException` ("Nullable object must have a value") because the custom converter returns `null` for unknown values, but the generated model code attempts to access `.Value` on the non-nullable enum property.

This is a regression/inconsistency compared to inline enums (nested inside models), which use the `IEnum` class pattern and handle unknown values gracefully by wrapping them.

This issue also addresses the root cause of #1612, which reported `NullReferenceException` on .NET 8 for similar scenarios.

### Root Cause
The current SDK generation uses the default OpenAPI generator for standalone enums, which produces standard C# `enum` types. These types cannot hold arbitrary string values. The generated `JsonConverter` for these enums is designed to throw an exception or return `null` when an unknown string is encountered.

In contrast, "inline" enums use a custom mustache template (`modelInnerEnum.mustache`) that generates an `IEnum` wrapper class. This class is designed to preserve unknown string values, making the SDK forward-compatible.

### Solution
Align the generation of standalone enums with the `IEnum` pattern used for inline enums.
1. Create/Override the `modelEnum.mustache` template to generate standalone enums as `IEnum` classes rather than standard C# `enum` types.
2. Update the generation logic to ensure all standalone enums use this forward-compatible pattern.

This will ensure that adding new values to the API does not break existing SDK integrations, and provides a consistent developer experience across all enum types in the library.

gcatanese:
## Description 

Fix #1650 #1612

### Problem 
There are 2 types of enum values in the .NET library:

- `inner enums`: they define an `IEnum` class that it is forward-compatible, new (unknown) values are handled
- `standalone enums`: they define standalone `public enums` (with int backing values), new (unknown) values throw an exception

### Solution 
Replaces all standalone C# enum types generated by `modelEnum.mustache` with an `IEnum` class pattern, matching the existing inner-enum approach. Unknown API values are now preserved as-is instead of throwing `JsonException` or causing `NullReferenceException` (fixes #1612).

The PR updates the Mustache templates and generates Checkout models (to create additional tests for the different type of enums) and BalancePlatform models (to fix a failing test).

> [!IMPORTANT]
>  👍 Handling of **standalone enum values** has been improved: new enum values do not thrown an error, they instead retain the string value found in the JSON payload. This was the goal of the PR.
>
> ⚠️ However, a similar change has been applied to **inline enum values**: unknown enum values are no longer deserialized as `null`, but they instead retain the string value found in the JSON payload.
> This is an improvement but also a breaking change: existing code that relies on null checks to identify unknown enum values will no longer work.

### Changes

- `modelEnum.mustache`: override default template and generate `IEnum` classes for standalone enums with `FromStringOrDefault`, `ToJsonValue`, implicit string conversions, and a nested `JsonConverter`
- modelInnerEnum.mustache: ToJsonValue returns raw string for unknown values instead of null
- JsonConverter.mustache: read path preserves unknown enum values; required enum assignment removes .Value unwrap; write path uses T.ToJsonValue(); oneOf discriminator resolution guards against unknown type values via FromStringOrDefault
- HostConfiguration.mustache: converter registration updated to nested class form; NullableJsonConverter removed
- modelGeneric.mustache: discriminator init uses implicit string cast instead of Enum.Parse
- ClientUtils.mustache: ToJsonValue call updated for standalone enums

### Tested scenarios
- StandaloneEnumTest.cs — 19 new tests covering the standalone IEnum pattern and GitHub #1612 regression
- IEnumCheckoutTest.cs — 22 new tests covering `Result` (standalone) and `PixPayByBankDetails` (inner TypeEnum) with real generated classes

### Affected Standalone Enums (18 files)
18 standalone enums listed above change from `enum` (value type) to `class : IEnum` (reference type).

| Module | Enums |
|---|---|
| BalancePlatform | `LimitStatus`, `AssociationStatus`, `Scope`, `ScaDeviceType`, `ScaExemption`, `MandateStatus`, `MandateType`, `SettingType`, `ScaStatus`, `ScaEntityType`, `TransferType` |
| Capital | `CALocalBankAccountType`, `AdditionalBankIdentificationTypes`, `FundsCollectionType`, `USLocalBankAccountType` |
| Checkout | `Result` |
| SessionAuthentication | `ResourceType`, `ProductType` |

**BEFORE (current SDK):**
```csharp
// Type is a C# enum (value type)
LimitStatus status = LimitStatus.Active;

// Switch statement
switch (status)
{
    case LimitStatus.Active:
        Console.WriteLine("Active");
        break;
    case LimitStatus.Inactive:
        Console.WriteLine("Inactive");
        break;
}

// Nullable handling (Nullable<T> value type)
LimitStatus? optional = GetStatus();
if (optional.HasValue)
{
    LimitStatus unwrapped = optional.Value;
}

// Default value
LimitStatus def = default; // = 0 (undefined int-backed value)

// Cast to int
int numeric = (int)status; // = 1

// Crashes on unknown API values:
// JsonException or NullReferenceException (.NET 8)
```

**AFTER (new SDK):**
```csharp
// Type is now an IEnum class (reference type)
LimitStatus status = LimitStatus.Active;

// Switch statement -> if/else or pattern matching
if (status == LimitStatus.Active)
    Console.WriteLine("Active");
else if (status == LimitStatus.Inactive)
    Console.WriteLine("Inactive");

// Nullable handling (standard null check)
LimitStatus? optional = GetStatus();
if (optional != null)
{
    LimitStatus unwrapped = optional; // no .Value needed
}

// Default value
LimitStatus def = default; // = null (reference type)

// Cast to int no longer works
// int numeric = (int)status; // COMPILE ERROR

// String value access
string raw = status.Value; // "active"

// Unknown API values are preserved (no crash):
LimitStatus unknown = (LimitStatus)"future-status";
Console.WriteLine(unknown.Value); // "future-status"
Console.WriteLine(unknown == LimitStatus.Active); // false
```

### Summary of breaking changes

| Change | Impact | Migration |
|---|---|---|
| `enum` -> `class` | `switch`/`case` no longer works | Use `if`/`==` or pattern matching |
| Value type -> Reference type | `.HasValue`/`.Value` on nullable removed | Use `!= null` instead of `.HasValue` |
| `default` changes from `0` to `null` | Variables default to null | Initialize explicitly |
| `(int)` cast removed | Numeric conversion fails | Use `.Value` string property |
| `Enum.Parse`/`Enum.GetValues` removed | Reflection-based enum APIs fail | Use `FromStringOrDefault` / static fields |


gcatanese:
@gemini-code-assist review

templates-v7/csharp/libraries/generichost/modelEnum.mustache
```diff
@@ -0,0 +1,125 @@
...
+    {{/description}}
+    {{#isString}}
+    [JsonConverter(typeof({{datatypeWithEnum}}{{^datatypeWithEnum}}{{classname}}{{/datatypeWithEnum}}.{{datatypeWithEnum}}{{^datatypeWithEnum}}{{classname}}{{/datatypeWithEnum}}JsonConverter))]
+    {{/isString}}
+    {{>visibility}} class {{datatypeWithEnum}}{{^datatypeWithEnum}}{{classname}}{{/datatypeWithEnum}} : IEnum
+    {
+        /// <summary>
+        /// Returns the value of the {{datatypeWithEnum}}{{^datatypeWithEnum}}{{classname}}{{/datatypeWithEnum}}.
+        /// </summary>
+        public string? Value { get; set; }
```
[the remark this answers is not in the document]
gcatanese: This is acceptable

Adyen.Test/Checkout/IEnumCheckoutTest.cs
```diff
@@ -0,0 +1,254 @@
...
+            Exception? ex = null;
+            Result? result = null;
+            try { result = JsonSerializer.Deserialize<Result>("\"NEW_RESULT_TYPE\"", _jsonSerializerOptionsProvider.Options); }
+            catch (Exception e) { ex = e; }
+            Assert.IsNull(ex);
+            Assert.IsNotNull(result);
+        }
+
+        [TestMethod]
+        public void Given_Result_When_DeserializeNull_Then_ReturnsNull()
```
galesky-a, line 78: Is this the expected behavior ? I think it may make sense but just double checking.

galesky-a (approved):
LGTM, best to tackle those asap so we have the same pattern for all enums moving forward. Left a single comment on `null` behavior but it's just a sanity check

Adyen.Test/Checkout/IEnumCheckoutTest.cs
gcatanese: In this specific helper method the expectation is correct

Issue #1687, opened 2026-05-15
[Bug] byte[] properties deserialize differently in bare STJ mode vs SDK pipeline


gcatanese:
## Context

The Adyen .NET SDK uses a custom `ByteArrayConverter` (in `Adyen.Core.Converters`) to serialize and deserialize `byte[]` properties. This converter treats byte arrays as raw UTF-8 strings in JSON, rather than the base64 encoding that `System.Text.Json` (STJ) uses by default.

The UTF-8 approach in `ByteArrayConverter` **is intentional and consistent with the Adyen Java SDK**, which applies the same treatment via `ByteArrayDeserializer` / `ByteArraySerializer` (registered globally on the Jackson `ObjectMapper` in each module's `JSON.java`). The semantics of `ByteArrayConverter` are therefore not in question — the fix is purely about aligning the bare STJ code path with the existing, correct cross-SDK behavior.

When models are deserialized through the SDK's normal HTTP pipeline (`JsonSerializerOptionsProvider`), `ByteArrayConverter` is registered in the `JsonSerializerOptions` and is used correctly. However, when a user deserializes a model using a bare `JsonSerializer.Deserialize<T>(json)` call — without supplying options — STJ falls back to its built-in `byte[]` handler, which base64-decodes the JSON string instead of treating it as UTF-8.

This means `byte[]` properties like `ThreeDSecureData.Cavv`, `ThreeDSecureData.Xid`, and `ThreeDSecureData.TokenAuthenticationVerificationValue` will be populated with different byte values depending on which deserialization path is used.

This issue was discovered while adding support for bare STJ deserialization of Adyen models (via class-level `[JsonConverter]` attributes). The class-level attribute correctly causes STJ to invoke the generated `XxxJsonConverter` in bare mode, but that converter internally calls `JsonSerializer.Deserialize<byte[]>(ref reader, jsonSerializerOptions)` with the passed-in (empty) options, bypassing `ByteArrayConverter` entirely.

The property-level `[JsonConverter(typeof(ByteArrayConverter))]` attribute added to `byte[]` properties does not help here, because it is only consulted by STJ during its own property-level deserialization — not when called explicitly from within a custom converter.

## Proposed Change

[...]

PR #1688, opened 2026-05-15
Support Bare deserialization and update tests

### LINKED ISSUE #1613: [Feature Request]: Support bare JsonSerializer.Deserialize<T>(json) without requiring SDK-internal JsonSerializerOptions
## Context

Since v34.0.0, the library uses `System.Text.Json` with per-model custom `JsonConverter<T>` instances, registered in each API namespace's `HostConfiguration`. The SDK's HTTP pipeline (`TryDeserializeOkResponse`, `Ok()`, etc.) passes these options automatically, so deserialization works correctly.

However, calling `JsonSerializer.Deserialize<T>(json)` **without the SDK's options** can produce incorrect results for certain field types.

## Current recommended approach

Use the SDK pipeline or explicitly pass `JsonSerializerOptionsProvider.Options` when deserializing outside the pipeline:

```csharp
// Option 1: Use the SDK pipeline (recommended)
var apiResponse = await checkout.PaymentMethodsAsync(request);
apiResponse.TryDeserializeOkResponse(out PaymentMethodsResponse response);

// Option 2: Retrieve the SDK options from DI and pass them explicitly
var provider = host.Services.GetRequiredService<JsonSerializerOptionsProvider>();
var result = JsonSerializer.Deserialize<PaymentMethodsResponse>(json, provider.Options);
```

Both approaches work correctly because the SDK's `JsonSerializerOptions` includes all the necessary converters for date formatting, `Option<T>` handling, enum serialization, polymorphic types, and per-model deserialization logic.

## Current workaround

For users who need to deserialize SDK models outside the DI/pipeline context today, the workaround is to manually build `JsonSerializerOptions` with the required converters:

```csharp
var options = new JsonSerializerOptions();
options.Converters.Add(new JsonStringEnumConverter());
options.Converters.Add(new DateTimeJsonConverter());
options.Converters.Add(new DateTimeNullableJsonConverter());
options.Converters.Add(new DateOnlyJsonConverter());
options.Converters.Add(new DateOnlyNullableJsonConverter());
options.Converters.Add(new ByteArrayConverter());
// Add each per-model converter you need, e.g.:
options.Converters.Add(new PaymentMethodsResponseJsonConverter());
options.Converters.Add(new PaymentMethodJsonConverter());
options.Converters.Add(new StoredPaymentMethodJsonConverter());
// ... etc.

var result = JsonSerializer.Deserialize<PaymentMethodsResponse>(json, options);
```

This is verbose and fragile (you must know which converters to register for the full object graph), but it works today without any library changes.

## Problem statement

Models are public types shipped in a NuGet package, so users reasonably expect `JsonSerializer.Deserialize<T>(json)` to work. While basic deserialization may succeed for simple models (the `[JsonConstructor]` / `Option<T>` binding issue from #1474 has been resolved across all models), **bare deserialization can still produce incorrect results** because:

1. **Custom date/byte/enum converters only registered in `HostConfiguration`.** The converters for `DateTime`, `DateOnly`, `byte[]`, and custom enum types are added to `JsonSerializerOptions` inside `HostConfiguration`, not attached to the types or properties themselves via `[JsonConverter]` attributes. Without the SDK's options, STJ uses its defaults, which may not match the API's date formats.

2. **Per-model `JsonConverter<T>` not discoverable by STJ.** Each model has a companion converter (e.g., `PaymentMethodsResponseJsonConverter`) that handles `Option<T>` wrapping and per-property deserialization. These are registered in `HostConfiguration` but not attached to the model class via a `[JsonConverter]` attribute, so bare STJ does not discover them. This means optional-field tracking (`Option<T>`) does not work correctly, and serialization may include fields that should be omitted (or omit fields that should be included).

**Affected scenarios include:** unit tests, caching layers, message queues, webhook payloads read from logs, and any context where models are deserialized outside the SDK's HTTP pipeline.

## Proposed solution

A **templates-only change** (no runtime architecture change) that makes bare `JsonSerializer.Deserialize<T>(json)` work correctly:

1. Add `[JsonConverter(typeof(XJsonConverter))]` at the **type level** on every generated model class. STJ discovers type-level converter attributes automatically, regardless of which `JsonSerializerOptions` the caller passes.
2. Add **property-level** `[JsonConverter]` attributes on `DateTime`, `DateTime?`, `DateOnly`, `DateOnly?`, and `byte[]` properties, pointing to the existing custom converters.
3. **Keep `HostConfiguration` unchanged.** Attribute-resolved converters and collection-registered converters are compatible in STJ; attribute-level takes precedence for the same type. The SDK pipeline continues to work exactly as before.

Since all model code is auto-generated from Mustache templates (`templates-v7/csharp/libraries/generichost/`), the change would be made in the templates and applied consistently on the next generation run.

### For the library

- **Polymorphic / oneOf models:** Nested types used in polymorphic discriminators (e.g., `CheckoutPaymentMethod`, `PaymentResponseAction`) must also carry type-level `[JsonConverter]` so the full object graph resolves correctly during bare deserialization.
- **Property-level date converters lock the format into the model.** If the API ever changes date format for a specific field, it would require a per-property override rather than a global options change. This is already the case via the existing `DateTimeJsonConverter`, but it becomes more tightly coupled.

### For users of the library

- **No breaking changes expected** for users of `TryDeserializeOkResponse` or the DI-based pipeline. Those paths continue to work identically.
- Users who previously passed custom `JsonSerializerOptions` **without** the SDK converters would now get different (correct) behavior for these types, because the type-level `[JsonConverter]` attribute takes precedence. This is strictly an improvement, but would be called out in release notes.
- Bare `JsonSerializer.Deserialize<T>(json)` would become a **supported** deserialization path for response models.

## We want your feedback

Before implementing this, we'd like to hear from the community:

1. **Do you need bare `JsonSerializer.Deserialize<T>(json)`?** If so, what are your use cases? (unit tests, caching, message queues, webhook log replay, something else?)
2. **Would the proposed approach** (type-level `[JsonConverter]` attributes on models) **meet your needs?**
3. **Do you have any concerns** about this change or the risks described above?

Please share your input in the comments below. If this gets enough traction, we will prioritize it for a future release.

gcatanese:
## Description

Fixes #1613 #1687

This PR implements the proposed solution from issue #1613: Support bare `JsonSerializer.Deserialize<T>(json)` without requiring SDK-internal `JsonSerializerOptions`.

### Problem

Since v34.0.0, custom `JsonConverter<T>` instances for each model were registered only in `HostConfiguration` via DI. This meant that calling `JsonSerializer.Deserialize<T>(json)` **without** the SDK's options produced incorrect results — missing optional-field tracking, wrong date formats, and broken byte array handling. Affected scenarios include unit tests, webhook processing, caching layers, and message queues.

### Solution

Made all generated models self-describing to `System.Text.Json` by adding `[JsonConverter(typeof(XJsonConverter))]` attributes at the **class level**, so STJ discovers the correct converter automatically regardless of which `JsonSerializerOptions` (if any) the caller supplies. Property-level `[JsonConverter]` attributes are also added for `byte[]` and `DateOnly` fields.

### Changes

- **`templates-v7/csharp/libraries/generichost/modelGeneric.mustache`** — Added class-level `[JsonConverter]` attribute and property-level attributes for `byte[]` and `DateOnly` types to the model template.
- **`templates-v7/csharp/libraries/generichost/model.mustache`** — Added missing `using` for `Adyen.Core.Converters` namespace.
- **models**: generate models for `Checkout`, `LEM`, `Capital`
- **`Adyen.Test/Core/Serialization/BareSerializationTests.cs`** — New test class covering bare serialize/deserialize round-trips for `Amount`, `ThreeDSecureData`, `CheckoutPaymentMethod`, and nested models without any `JsonSerializerOptions`.
- **tests in various services**: add tests for bare deserialization in Checkout, LEM, Capital 
- **`Adyen.Test/Payment/PaymentTest.cs`** — Fixed existing test assertion: use `TryGetProperty` instead of `GetProperty` for optional non-nullable fields that may be absent in the serialized output.
- **`README.md`** — Documented the new bare deserialization path and clarified when DI-provided options are still recommended (e.g. for `byte[]` fields with `ByteArrayConverter`).

### Backwards compatibility

No breaking changes. STJ gives precedence to type-level `[JsonConverter]` attributes over converters registered in `JsonSerializerOptions`, so the existing DI-based pipeline continues to work exactly as before.

### Known limitation

`byte[]` fields (e.g. `ThreeDSecureData.Cavv`) behave differently in bare mode vs. DI mode due to how `ByteArrayConverter` is currently implemented. This is documented in the README and tracked separately in #1687.


Adyen/Checkout/Models/ThreeDSecureData.cs
```diff
@@ -584,6 +586,7 @@ public override void Write(Utf8JsonWriter writer, DirectoryResponseEnum value, J
         /// </summary>
         /// <value>The cardholder authentication value (base64 encoded, 20 bytes in a decoded form).</value>
         [JsonPropertyName("cavv")]
+        [JsonConverter(typeof(ByteArrayConverter))]
```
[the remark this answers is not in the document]
gcatanese: Removed in the templates, will be fixed after the next code generation

templates-v7/csharp/libraries/generichost/modelGeneric.mustache
```diff
@@ -214,6 +215,11 @@
         /* <example>{{.}}</example> */
         {{/example}}
         [JsonPropertyName("{{baseName}}")]
+        {{#isDate}}
+        {{#required}}[JsonConverter(typeof(DateOnlyJsonConverter))]{{/required}}{{^required}}[JsonConverter(typeof(DateOnlyNullableJsonConverter))]{{/required}}
```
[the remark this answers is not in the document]
gcatanese: Corrected

templates-v7/csharp/libraries/generichost/JsonConverter.mustache
```diff
@@ -631,14 +636,14 @@
             if ({{#lambda.camelcase_sanitize_param}}{{classname}}{{/lambda.camelcase_sanitize_param}}.{{name}} != null)
             {
                 writer.WritePropertyName("{{baseName}}");
-                JsonSerializer.Serialize(writer, {{#lambda.camelcase_sanitize_param}}{{classname}}{{/lambda.camelcase_sanitize_param}}.{{name}}, jsonSerializerOptions);
+                {{#isByteArray}}new ByteArrayConverter().Write(writer, {{#lambda.camelcase_sanitize_param}}{{classname}}{{/lambda.camelcase_sanitize_param}}.{{name}}, jsonSerializerOptions);{{/isByteArray}}{{^isByteArray}}JsonSerializer.Serialize(writer, {{#lambda.camelcase_sanitize_param}}{{classname}}{{/lambda.camelcase_sanitize_param}}.{{name}}, jsonSerializerOptions);{{/isByteArray}}
```
Kwok-he-Chu, line 639: If you're moving forward with the idea of doing deserialization through global attributes, I recommend taking the approach of adding the ByteConverters to the fields themselves.

Here's an example w/ `GetTermsOfServiceDocumentResponse.cs` and the ByteArray attribute*


```csharp
[JsonPropertyName("document")]
[JsonConverter(typeof(ByteArrayConverter))] //  You can update the mustache template accordingly {{#isByteArray}} ... add this line instead {{/isByteArray}}
public byte[]? Document
{
    get { return this._DocumentOption; }
    set { this._DocumentOption = new(value); }
}
```

README.md
```diff
@@ -167,6 +168,19 @@ var paymentRequest = JsonSerializer.Deserialize<PaymentRequest>("YOUR_JSON_STRIN
 var paymentResponse = new PaymentResponse(); // Replace with your actual PaymentResponse object
 var json = JsonSerializer.Serialize(paymentResponse, jsonSerializerOptionsProvider.Options);
```
Kwok-he-Chu, line 169: nit: `jsonSerializerOptionsProvider.Options` can be removed here (from the README)

templates-v7/csharp/libraries/generichost/modelGeneric.mustache
```diff
@@ -1,6 +1,7 @@
     /// <summary>
     /// {{{description}}}{{^description}}{{classname}}{{/description}}.
     /// </summary>
+    [JsonConverter(typeof({{classname}}JsonConverter))]
```
Kwok-he-Chu, line 4: With this change, (if you do move forward with attributing each field, f.e. ByteArray), should we also remove these injected Converters from `HostConfiguration.mustache`? https://github.com/Adyen/adyen-dotnet-api-library/blob/v34.0.3/templates-v7/csharp/libraries/generichost/HostConfiguration.mustache#L66

They will be injected twice now (one globally, and one in the specific JsonSerializerOptionsProvider instance), modifying the JsonSerializerOptionsProvider **will not affect the global one**

O

Kwok-he-Chu, line 4: Looking at `HostConfiguration.mustache` (https://github.com/Adyen/adyen-dotnet-api-library/blob/v34.0.3/templates-v7/csharp/libraries/generichost/HostConfiguration.mustache#L44-L69), I recommend putting the attributes on the models as well, similar to `ByteArrayConverter()` above.

example
``` 
            if (_jsonOptions.Converters.FirstOrDefault(x => x.GetType() == typeof(JsonStringEnumConverter)) == null)
                _jsonOptions.Converters.Add(new JsonStringEnumConverter());
            if (_jsonOptions.Converters.FirstOrDefault(x => x.GetType() == typeof(DateTimeJsonConverter)) == null)
                _jsonOptions.Converters.Add(new DateTimeJsonConverter());
            if (_jsonOptions.Converters.FirstOrDefault(x => x.GetType() == typeof(DateTimeNullableJsonConverter)) == null)
                _jsonOptions.Converters.Add(new DateTimeNullableJsonConverter());
            if (_jsonOptions.Converters.FirstOrDefault(x => x.GetType() == typeof(ByteArrayConverter)) == null)
                _jsonOptions.Converters.Add(new ByteArrayConverter());
```

Kwok-he-Chu (changes requested):
I've reviewed the changes, mainly addressing the attribute-usage in `HostConfiguration.cs` and on top of the models (classes).


FYI -- Had to change one more thing to make the TESTs compile in:


> Adyen.Test/Checkout/CheckoutPaymentsBaseSerilizationTests and PaymentsTest

Before:
```csharp
Action = new CheckoutThreeDS2Action
{
    Type = CheckoutThreeDS2Action.TypeEnum.ThreeDS2,
    PaymentMethodType = "scheme",
    PaymentData = "test-payment-data",
},
```

After:
```csharp
Action = new PaymentDetailsResponseAction(new CheckoutThreeDS2Action
{
    Type = CheckoutThreeDS2Action.TypeEnum.ThreeDS2,
    PaymentMethodType = "scheme",
    PaymentData = "test-payment-data"
})
```

templates-v7/csharp/libraries/generichost/JsonConverter.mustache
gcatanese: Code compile and test both locally and on GitHub, check if the branch is not in sync or you need to rebuild the package?

The class-level `[JsonConverter(typeof(ThreeDSecureDataJsonConverter))]` means the custom `ThreeDSecureDataJsonConverter.Read/Write` methods take full control of serialization. They already handle `byte[]` fields correctly by directly instantiating new ByteArrayConverter().Read/Write(...) (lines 783,
    886, etc.). 
I understand property-level [JsonConverter] attributes are completely ignored by STJ when a class-level converter exists.

README.md
gcatanese: This shows an example of  DI-based deserialization (recommended approach), which intentionally shows how to use the provider's options. 

The bare deserialization section below instead shows usage without options. The two sections serve different purposes.

templates-v7/csharp/libraries/generichost/modelGeneric.mustache
gcatanese: Same comment as https://github.com/Adyen/adyen-dotnet-api-library/pull/1688#discussion_r3319616719

gcatanese: SeeCorrect, but harmless since the options-level registrations are never reached for these models. I have created an issue for a future release. #1728

templates-v7/csharp/libraries/generichost/JsonConverter.mustache
Kwok-he-Chu: Ah, you're right. *facepalm*

When I reviewed this model -> https://github.com/Adyen/adyen-dotnet-api-library/blob/feature/1613-support-base-jst/Adyen/Payment/Models/ThreeDSecureData.cs#L812-L815, I noticed that it **did not use the ByteArray** converter. 

```csharp
///   public override ThreeDSecureData Read( ....
/// ...
     case "xid":
                            xid = new Option<byte[]?>(JsonSerializer.Deserialize<byte[]>(ref utf8JsonReader, jsonSerializerOptions)!);
                            break;
/// ...
```

This is only true if the JsonSerializerOptions **registers the converter** in HostConfiguration (https://github.com/Adyen/adyen-dotnet-api-library/blob/feature/1613-support-base-jst/Adyen/Payment/Client/HostConfiguration.cs#L55-L56)

```
if (_jsonOptions.Converters.FirstOrDefault(x => x.GetType() == typeof(ByteArrayConverter)) == null)
                _jsonOptions.Converters.Add(new ByteArrayConverter());
```

Can we confirm that, when HostConfiguration.cs does **not** register the ByteArrayConverter, will serialization/deserialization still behave similarly? Idem. for other converters*

gcatanese: Makes sense.
BTW Payment `ThreeDSecureData.cs` wasn't regenerated in this PR so it was still based on the old template.

I've updated it manually (added the class-level `[JsonConverter]` attribute, etc..) and added a bunch of bare (de)serialization tests for this model. The remaining (stale) models will be fixed when they are next regenerated by SDK Automation Bot.

Adyen/Payment/Models/ThreeDSecureData.cs
```diff
@@ -778,7 +780,7 @@ public override ThreeDSecureData Read(ref Utf8JsonReader utf8JsonReader, Type ty
                             authenticationResponse = new Option<ThreeDSecureData.AuthenticationResponseEnum?>(ThreeDSecureData.AuthenticationResponseEnum.FromStringOrDefault(authenticationResponseRawValue) ?? (ThreeDSecureData.AuthenticationResponseEnum)authenticationResponseRawValue);
                             break;
                         case "cavv":
-                            cavv = new Option<byte[]?>(JsonSerializer.Deserialize<byte[]>(ref utf8JsonReader, jsonSerializerOptions)!);
+                            cavv = new Option<byte[]?>(new ByteArrayConverter().Read(ref utf8JsonReader, typeof(byte[]), jsonSerializerOptions));
```
Kwok-he-Chu, line 783: I see two main ways to handle this `byte[]` deserialization. Ideally, the SDK should keep using the following line, allowing consumers to register their own byte[] converters passed to the `JsonSerializerOptions`.

____

**Option 1:** Consumer-facing registration
```csharp
// consumer:
services.AddControllers()
    .AddJsonOptions(options =>
    {
        options.JsonSerializerOptions.Converters.Add(new ByteArrayConverter()); // requires consumer to specify the converter
    });

// SDK:
case "cavv":
    cavv = new Option<byte[]?>(
        JsonSerializer.Deserialize<byte[]>(
            ref utf8JsonReader,
            jsonSerializerOptions)); // preferred way and use: byte[]
    break;
```


**Option 2:** Do an explicit look-up, and check if if we're using System.Text.Json or our own. We can enforce it here // note: pseudo code, needs to be tested*
_Example:_
```csharp
/// SDK
bool isCustomConverter =
    options.Converters.Any(c => c.CanConvert(typeof(byte[])));             // if it's false we use "System.Text.Json.Serialization.Converters.ByteArrayConverter"  , not sure if I have to check the fully-qualified name here

// Then assign or instantiate the right converter
i
              JsonConverter<byte[]> byteArrayConverter;

if (!isCustomConverter)
    byteArrayConverter = (JsonConverter<byte[]>)jsonSerializerOptions.GetConverter(typeof(byte[]));
else
    byteArrayConverter = new Adyen.Core.Converters.ByteArrayConverter();

case "cavv":
    cavv = new Option<byte[]?>(
        byteArrayConverter.Read(
            ref utf8JsonReader,
            typeof(byte[]),
            jsonSerializerOptions));
    break;
```

Adyen/Payment/Models/ThreeDSecureData.cs
```diff
@@ -918,7 +920,7 @@ public void WriteProperties(Utf8JsonWriter writer, ThreeDSecureData threeDSecure
             if (threeDSecureData._TokenAuthenticationVerificationValueOption.IsSet)
             {
                 writer.WritePropertyName("tokenAuthenticationVerificationValue");
-                JsonSerializer.Serialize(writer, threeDSecureData.TokenAuthenticationVerificationValue, jsonSerializerOptions);
+                new ByteArrayConverter().Write(writer, threeDSecureData.TokenAuthenticationVerificationValue, jsonSerializerOptions);
```
Kwok-he-Chu, line 923: nit:  This instance of `new ByteArrayConverter().Read(...) ` gets created every time. Perhaps we can do something like:


```csharp
// example
public class ThreeDSecureDataJsonConverter : JsonConverter<ThreeDSecureData> { 
       private readonly ByteArrayConverter byteArrayConverter;
       public ThreeDSecureDataJsonConverter()
       {
                byteArrayConverter = new ByteArrayConverter();
        }
}
```

Adyen/Payment/Models/ThreeDSecureData.cs
```diff
@@ -881,7 +883,7 @@ public void WriteProperties(Utf8JsonWriter writer, ThreeDSecureData threeDSecure
             if (threeDSecureData._CavvOption.IsSet)
             {
                 writer.WritePropertyName("cavv");
-                JsonSerializer.Serialize(writer, threeDSecureData.Cavv, jsonSerializerOptions);
+                new ByteArrayConverter().Write(writer, threeDSecureData.Cavv, jsonSerializerOptions);
```
Kwok-he-Chu, line 886: https://github.com/Adyen/adyen-dotnet-api-library/pull/1688/changes#r3324189554

Adyen/Payment/Models/ThreeDSecureData.cs
```diff
@@ -804,13 +806,13 @@ public override ThreeDSecureData Read(ref Utf8JsonReader utf8JsonReader, Type ty
...
                         case "tokenAuthenticationVerificationValue":
-                            tokenAuthenticationVerificationValue = new Option<byte[]?>(JsonSerializer.Deserialize<byte[]>(ref utf8JsonReader, jsonSerializerOptions)!);
+                            tokenAuthenticationVerificationValue = new Option<byte[]?>(new ByteArrayConverter().Read(ref utf8JsonReader, typeof(byte[]), jsonSerializerOptions));
                             break;
                         case "transStatusReason":
                             transStatusReason = new Option<string?>(utf8JsonReader.GetString()!);
                             break;
                         case "xid":
-                            xid = new Option<byte[]?>(JsonSerializer.Deserialize<byte[]>(ref utf8JsonReader, jsonSerializerOptions)!);
+                            xid = new Option<byte[]?>(new ByteArrayConverter().Read(ref utf8JsonReader, typeof(byte[]), jsonSerializerOptions));
```
Kwok-he-Chu, line 815: https://github.com/Adyen/adyen-dotnet-api-library/pull/1688/changes#r3324189554

Adyen/Payment/Models/ThreeDSecureData.cs
Kwok-he-Chu, line 783: https://github.com/Adyen/adyen-dotnet-api-library/pull/1688/changes#r3324189554

Adyen/Payment/Models/ThreeDSecureData.cs
```diff
@@ -927,7 +929,7 @@ public void WriteProperties(Utf8JsonWriter writer, ThreeDSecureData threeDSecure
             if (threeDSecureData._XidOption.IsSet)
             {
                 writer.WritePropertyName("xid");
-                JsonSerializer.Serialize(writer, threeDSecureData.Xid, jsonSerializerOptions);
+                new ByteArrayConverter().Write(writer, threeDSecureData.Xid, jsonSerializerOptions);
```
Kwok-he-Chu, line 932: https://github.com/Adyen/adyen-dotnet-api-library/pull/1688/changes#r3324189554

Kwok-he-Chu (changes requested):
Review pass 2 - for the ByteArrayConverter

Adyen/Payment/Models/ThreeDSecureData.cs
gcatanese: Created #1738

Adyen/Payment/Models/ThreeDSecureData.cs
gcatanese: Created #1739

Kwok-he-Chu (approved):

PR #1691, opened 2026-05-19
fix: return null for unknown oneOf discriminator types to ensure forward compatibility

### LINKED ISSUE #1690: [Bug] Forward Compatibility not honored: oneOf models throw JsonException on unknown types
### Description
Current implementation of `oneOf` models (composition-based, like `CheckoutPaymentMethod` and `PaymentResponseAction`) throws a `JsonException` during deserialization if the `type` discriminator is unknown.

### Problem
Adyen frequently adds new payment methods and action types to the API. When a new type is introduced, existing versions of the .NET SDK will fail to deserialize any response containing the new type. This turns an additive API change into a breaking change for merchants using older SDK versions, as the entire response fails to deserialize.

As verified in PR #1689, models using inheritance (like `ShopperIdPaymentMethod`) handle this gracefully by falling back to the base class, but models using composition (`oneOf`) fail strictly.

### Expected Behavior
The SDK should handle unknown types in `oneOf` models gracefully to ensure forward compatibility. Possible approaches:
- Map unknown types to a generic fallback type.
- Return `null` for the specific field instead of failing the entire deserialization.
- Adjust Mustache templates to provide a catch-all for unknown discriminators in composition models.

### Verification
The behavior is confirmed by the following tests in `Adyen.Test/Checkout/PaymentsTest.cs`:
- `Given_CheckoutPaymentMethod_With_Unknown_Type_When_Deserialized_Then_Throws_JsonException`
- `Given_PaymentResponseAction_With_Unknown_Type_When_Deserialized_Then_Throws_JsonException`

gcatanese:
## Description
Fixes #1690.

`oneOf` converters (e.g. `CheckoutPaymentMethod`, `PaymentResponseAction`) previously threw `JsonException` when an unknown `type` discriminator was encountered. This turned additive API changes (new payment methods / action types) into breaking changes for merchants on older SDK versions.

The fix replaces `throw new JsonException()` with `return null!` in the oneOf fallback path, so that an unknown type causes the field to be `null` rather than failing the entire response deserialization.

The Mustache template is updated so all future regenerated `oneOf` classes inherit the same behaviour.

## Changes
- `templates-v7/csharp/libraries/generichost/JsonConverter.mustache` – replace `throw new JsonException()` with `return null!` in the `oneOf` fallback (discriminator fallback is unchanged)
- `Adyen/Checkout/Models/CheckoutPaymentMethod.cs` – same fix applied to the generated class
- `Adyen/Checkout/Models/PaymentResponseAction.cs` – same fix applied to the generated class
- `Adyen.Test/Checkout/PaymentsTest.cs` – three new tests:
  - `Given_ShopperIdPaymentMethod_With_Unknown_Type_When_Deserialized_Then_ReturnsBaseInstance` (allOf / discriminator path – already worked, test documents the behaviour)
  - `Given_CheckoutPaymentMethod_With_Unknown_Type_When_Deserialized_Then_ReturnsNull`
  - `Given_PaymentResponseAction_With_Unknown_Type_When_Deserialized_Then_ReturnsNull`

## Tested scenarios
- All 558 existing + new unit tests pass (`dotnet test Adyen.Test`)
- Unknown type in `CheckoutPaymentMethod` → `null` (no exception)
- Unknown type in `PaymentResponseAction` → `null` (no exception)
- Unknown type in `ShopperIdPaymentMethod` → base instance (unchanged, documents existing behaviour)

## Fixed issue
Closes #1690

### DIFF
--- a/Adyen.Test/Checkout/PaymentsTest.cs
+++ b/Adyen.Test/Checkout/PaymentsTest.cs
@@ -1230,5 +1230,53 @@ namespace Adyen.Test.Checkout
         }
 
         #endregion
+
+        #region allOf unknown type
+
+        [TestMethod]
+        public void Given_ShopperIdPaymentMethod_With_Unknown_Type_When_Deserialized_Then_ReturnsBaseInstance()
+        {
+            // Arrange
+            string json = @"{""type"":""unknown_shopper_id_method"",""someField"":""someValue""}";
+
+            // Act
+            var result = JsonSerializer.Deserialize<ShopperIdPaymentMethod>(json, _jsonSerializerOptionsProvider.Options);
+
+            // Assert
+            Assert.IsNotNull(result);
+            Assert.AreEqual(typeof(ShopperIdPaymentMethod), result.GetType());
+        }
+
+        #endregion
+
+        #region oneOf unknown type
+
+        [TestMethod]
+        public void Given_CheckoutPaymentMethod_With_Unknown_Type_When_Deserialized_Then_ReturnsNull()
+        {
+            // Arrange
+            string json = @"{""type"":""unknown_payment_method"",""someField"":""someValue""}";
+
+            // Act
+            var result = JsonSerializer.Deserialize<CheckoutPaymentMethod>(json, _jsonSerializerOptionsProvider.Options);
+
+            // Assert
+            Assert.IsNull(result);
+        }
+
+        [TestMethod]
+        public void Given_PaymentResponseAction_With_Unknown_Type_When_Deserialized_Then_ReturnsNull()
+        {
+            // Arrange
+            string json = @"{""type"":""unknown_action"",""url"":""https://example.com""}";
+
+            // Act
+            var result = JsonSerializer.Deserialize<PaymentResponseAction>(json, _jsonSerializerOptionsProvider.Options);
+
+            // Assert
+            Assert.IsNull(result);
+        }
+
+        #endregion
     }
 }
--- a/Adyen/Checkout/Models/CheckoutPaymentMethod.cs
+++ b/Adyen/Checkout/Models/CheckoutPaymentMethod.cs
@@ -1498,7 +1498,7 @@ namespace Adyen.Checkout.Models
             if (zipDetails?.Type != null && ZipDetails.TypeEnum.FromStringOrDefault((string?)zipDetails.Type) != null)
                 return new CheckoutPaymentMethod(zipDetails);
 
-            throw new JsonException();
+            return null!;
         }
 
         /// <summary>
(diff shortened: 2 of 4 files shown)


Adyen.Test/Checkout/PaymentsTest.cs
```diff
@@ -1230,5 +1230,53 @@ public void Given_Serialize_When_PaymentDetailsResponse_WithThreeDS2Action_Then_
...
+        {
+            // Arrange
+            string json = @"{""type"":""unknown_shopper_id_method"",""someField"":""someValue""}";
+
+            // Act
+            var result = JsonSerializer.Deserialize<ShopperIdPaymentMethod>(json, _jsonSerializerOptionsProvider.Options);
+
+            // Assert
+            Assert.IsNotNull(result);
+            Assert.AreEqual(typeof(ShopperIdPaymentMethod), result.GetType());
```
Kwok-he-Chu, line 1247: Anchor

Adyen.Test/Checkout/PaymentsTest.cs
```diff
@@ -1230,5 +1230,53 @@ public void Given_Serialize_When_PaymentDetailsResponse_WithThreeDS2Action_Then_
...
+        #region oneOf unknown type
+
+        [TestMethod]
+        public void Given_CheckoutPaymentMethod_With_Unknown_Type_When_Deserialized_Then_ReturnsNull()
+        {
+            // Arrange
+            string json = @"{""type"":""unknown_payment_method"",""someField"":""someValue""}";
+
+            // Act
+            var result = JsonSerializer.Deserialize<CheckoutPaymentMethod>(json, _jsonSerializerOptionsProvider.Options);
```
Kwok-he-Chu, line 1261: I noticed a difference in `allOf` vs. `oneOf` unknown type behavior:

> A. `CheckoutPaymentMethod` is null here and does **not** preserve the `unknown_payment_method`-type

> B. While in the `allOf` example (https://github.com/Adyen/adyen-dotnet-api-library/pull/1691/changes#r3318851905) ->  result.Type does preserve its type `unknown_shopper_id_method`

Is this intended behavior?

Kwok-he-Chu, line 1261: **Mandatory** release notes: We need to explicitly state that this behavior is different in V35 ([before]: consumers need catch the `JsonException()`, [after]: now you need to check for null)

Kwok-he-Chu (commented):
Added 2 comments for this particular change

Adyen.Test/Checkout/PaymentsTest.cs
gcatanese: Sure

gcatanese: Yes (well spotted 😄 )
In `oneOf`, when the `type` is unknown, we don't know which payment method to choose, so we need to use `null` to make it explicit (no instance can be created)
In `allOf`, there is a safer (ish) fallback since we can defer to the parent class (preserving the value)

Kwok-he-Chu (approved):

PR #1718, opened 2026-05-21
AllOf without mapping not working

### LINKED ISSUE #1717: [Bug] Discriminator in allOf without mapping generates class name as wire value instead of API type
## Context

When an OpenAPI schema uses `allOf` inheritance with a discriminator that has an explicit `mapping`, the .NET SDK generator still serializes and deserializes using C# class names as the discriminator wire value instead of the mapped API wire values.

Affected model: `Resource` (SessionAuthentication API) — child classes `LegalEntityResource`, `AccountHolderResource`, `PaymentInstrumentResource`.

The OpenAPI spec **already has** an explicit mapping:

```yaml
Resource:
  discriminator:
    propertyName: type
    mapping:
      accountHolder: '#/components/schemas/AccountHolderResource'
      legalEntity: '#/components/schemas/LegalEntityResource'
      paymentInstrument: '#/components/schemas/PaymentInstrumentResource'
```

Yet the generated converter compares against class names:

```csharp
// ResourceJsonConverter.Read() — generated code (broken)
if (discriminator != null && discriminator.Equals("LegalEntityResource"))  // should be "legalEntity"
    return JsonSerializer.Deserialize<LegalEntityResource>(...);
```

And the constructor sets the wrong wire value:

```csharp
// Resource() constructor — generated code (broken)
Type = (ResourceType)this.GetType().Name;  // produces "LegalEntityResource" instead of "legalEntity"
```

The API sends `"type": "legalEntity"`, so deserialization silently returns a bare `Resource` instead of `LegalEntityResource`, and serialized requests carry `"type": "LegalEntityResource"` which the API rejects.

## Root Cause

The bug is in `JsonConverter.mustache`. The template has two separate discriminator paths:

1. **`{{#discriminator}}{{#children}}`** — used for `allOf` inheritance hierarchies. Uses `{{name}}` (C# class name) for matching. **This is the broken path.**
2. **`{{#model.hasDiscriminatorWithNonEmptyMapping}}` + `{{#mappedModels}}`** — uses `{{mappingName}}` (wire values). This path works correctly but is gated behind `{{#model.composedSchemas.oneOf}}`, so it **only fires for `oneOf` models**, never for `allOf` models like `Resource`.

The same issue exists in `modelGeneric.mustache` where the constructor always uses `this.GetType().Name` regardless of whether a discriminator mapping is present.

## Fix

Fixed in PR #1775. Changes to two templates:

- **`JsonConverter.mustache`**: Inside the `{{#discriminator}}` block, branch on `model.hasDiscriminatorWithNonEmptyMapping`. When true, use `{{#mappedModels}}`/`{{mappingName}}`/`{{model.classname}}` (wire values). When false, keep the original `{{#children}}`/`{{name}}` path — no impact on models without explicit mapping.
- **`modelGeneric.mustache`**: When `hasDiscriminatorWithNonEmptyMapping` is true, generate a switch expression that maps runtime class names to their correct wire values instead of using `this.GetType().Name`.

`Resource.cs` was manually updated to match the corrected template output.

gcatanese:
Fix #1717

WIP to address the handling of `allOf`

Note `Adyen.Test/mocks/sessionauthentication/create-session-request.json` type value should be `LegalEntity`

### DIFF
--- a/Adyen.Test/AcsWebhooks/AcsBareSerializationTests.cs
+++ b/Adyen.Test/AcsWebhooks/AcsBareSerializationTests.cs
@@ -1,0 +1,119 @@
+using System.Text.Json;
+using Adyen.AcsWebhooks.Models;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Adyen.Test.AcsWebhooks
+{
+    /// <summary>
+    /// Verifies that AcsWebhooks models can be deserialized using a bare
+    /// <see cref="JsonSerializer.Deserialize{T}(string)"/> call without supplying any
+    /// <see cref="JsonSerializerOptions"/>. Enum types carry class-level <c>[JsonConverter]</c>
+    /// attributes that STJ discovers automatically; plain model properties are resolved via
+    /// <c>[JsonPropertyName]</c> attributes through STJ's default reflection path.
+    /// </summary>
+    [TestClass]
+    public class AcsBareSerializationTests
+    {
+        [TestMethod]
+        public void Given_AuthenticationNotificationJson_When_BareDeserialize_Then_Returns_CorrectTopLevelFields()
+        {
+            string json = TestUtilities.GetTestFileContent("mocks/acswebhooks/balancePlatform.authentication.created.json");
+
+            var r = JsonSerializer.Deserialize<AuthenticationNotificationRequest>(json);
+
+            Assert.IsNotNull(r);
+            Assert.AreEqual(AuthenticationNotificationRequest.TypeEnum.BalancePlatformAuthenticationCreated, r.Type);
+            Assert.AreEqual("test", r.Environment);
+            Assert.AreEqual(DateTimeOffset.Parse("2022-12-22T15:42:03+01:00"), r.Timestamp);
+        }
+
+        [TestMethod]
+        public void Given_AuthenticationNotificationJson_When_BareDeserialize_Then_Returns_CorrectDataFields()
+        {
+            string json = TestUtilities.GetTestFileContent("mocks/acswebhooks/balancePlatform.authentication.created.json");
+
+            var r = JsonSerializer.Deserialize<AuthenticationNotificationRequest>(json);
+
+            Assert.IsNotNull(r?.Data);
+            Assert.AreEqual("497f6eca-6276-4993-bfeb-53cbbbba6f08", r.Data.Id);
+            Assert.AreEqual("YOUR_BALANCE_PLATFORM", r.Data.BalancePlatform);
+            Assert.AreEqual("PI3227C223222B5BPCMFXD2XG", r.Data.PaymentInstrumentId);
+            Assert.AreEqual(AuthenticationNotificationData.StatusEnum.Authenticated, r.Data.Status);
+        }
+
+        [TestMethod]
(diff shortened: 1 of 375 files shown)


gcatanese:
## Analysis of PR #1718 vs Issue #1717

### Root Cause

The bug is in `JsonConverter.mustache` — the `{{#discriminator}}{{#children}}` block uses `{{name}}` (C# class name) as the discriminator match value instead of the API wire value. Example from `Resource.cs`:

```csharp
// Generated: matches on class name — never matches real API responses
if (discriminator != null && discriminator.Equals("LegalEntityResource"))
```

The API sends `"type": "legalEntity"` (confirmed by the `ResourceType` enum), so the match always fails and a bare `Resource` is returned, silently losing all child-specific properties.

### Why the OpenAPI spec mapping is the correct fix

A template-only fix is **not feasible**: the `{{#children}}` block only has access to `{{name}}` (class name) and `{{classname}}`. The wire value (`{{mappingName}}`) is only available inside `{{#mappedModels}}`, which is empty when no explicit `discriminator.mapping` exists in the spec. The template simply lacks the data to connect `"legalEntity"` to `LegalEntityResource`.

Adding explicit `discriminator.mapping` to the OpenAPI spec activates the `{{#model.hasDiscriminatorWithNonEmptyMapping}}` code path, which correctly matches on `{{mappingName}}` (wire values). The `{{#children}}` block still generates but harmlessly misses (class names != wire values), and the mapping block handles it correctly. This is how all other discriminated models in the SDK already work.

### Affected models (5 total, not just the 2 in the issue)

| Model | Module | Children | Confirmed wire values |
|-------|--------|----------|-----------------------|
| `Resource` | SessionAuthentication | AccountHolderResource, LegalEntityResource, PaymentInstrumentResource | `accountHolder`, `legalEntity`, `paymentInstrument` (via `ResourceType` enum) |
| `ShopperIdPaymentMethod` | Checkout | PayToPaymentMethod, UPIPaymentMethod | Need verification |
| `BankAccountIdentification` | Capital | 16 children (AU/BR/CA/...Local) | Need verification |
| `MandateAccountIdentification` | BalancePlatform | UKLocalMandateAccountIdentification | Need verification |
| `WebhookSetting` | BalancePlatform | BalanceWebhookSetting | Need verification |

### Observations on this PR's scope

This PR conflates two separate concerns:

1. **The discriminator bug** (Issue #1717) — affects only the 5 models above
2. **Bare serialization support** — adding `[JsonConverter(typeof(...))]` attributes to ~95 Checkout model classes, `ByteArrayConverter` refactoring, and broad bare-serialization tests

The `[JsonConverter]` attribute changes, `ByteArrayConverter` refactoring, and bare-serialization test suite are unrelated to Issue #1717. This expands the change from ~5 files to ~100 files. Consider splitting these into separate PRs.

### Recommended approach

1. **Upstream fix**: Add explicit `discriminator.mapping` entries to the affected schemas in `adyen-openapi`, then regenerate. This is the only correct and sustainable fix.
2. **Short-term patch** (until mapping is available): Hand-edit the 5 affected generated files to fix the discriminator match strings, plus add focused discriminator deserialization tests per affected model. These tests act as regression guards — if regeneration happens before the mapping is added, the tests fail immediately.

PR #1727, opened 2026-05-28
Use byte comparison helper in HmacValidator.IsValidHmac

tonghuaroot:
## Summary

`HmacValidator.IsValidHmac` (Payments webhook signature verification) currently compares the computed and received signatures with `string.Equals`. The rest of the SDK already verifies HMAC signatures with a length-independent byte comparison in `Adyen.Core.Utilities.HmacValidatorUtility`. This change aligns `IsValidHmac` with that existing helper so both webhook signature paths compare signatures the same way.

## Changes

- `HmacValidator.IsValidHmac` now compares the UTF-8 bytes of the two signatures via a `TimeSafeEquals` helper that mirrors the one in `HmacValidatorUtility`.
- Adds an IL-inspection regression test (`TestIsValidHmac_UsesByteComparisonHelper`) asserting `IsValidHmac` does not call `String.Equals` / `String.op_Equality`.

All existing `HmacValidatorTest` cases continue to pass.

### DIFF
--- a/Adyen.Test/Webhooks/HmacValidatorTest.cs
+++ b/Adyen.Test/Webhooks/HmacValidatorTest.cs
@@ -153,53 +153,27 @@ namespace Adyen.Test.Webhooks
         }
 
         /// <summary>
-        /// IsValidHmac compares signatures with the SDK's length-independent
-        /// byte comparison helper, not plain String.Equals / String.op_Equality,
-        /// keeping it consistent with HmacValidatorUtility. Verified by IL
-        /// inspection of the compiled method.
+        /// A null hmacSignature value in AdditionalData must return false rather
+        /// than throwing — the comparison reads the signature value's bytes, so a
+        /// null value has to be guarded (the key can be present with a null value).
         /// </summary>
         [TestMethod]
-        public void TestIsValidHmac_UsesByteComparisonHelper()
+        public void TestIsValidHmac_NullSignatureValueReturnsFalse()
         {
-            var method = typeof(HmacValidator).GetMethod("IsValidHmac",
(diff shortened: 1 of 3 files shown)


gcatanese:
Thank you @tonghuaroot for your contribution, please address/reply/resolve to the Gemini comments (my review with Claude didn't flag anything to be honest, but  we want to make sure that all Gemini comments are noted)

tonghuaroot:
Thanks @gcatanese — addressed both Gemini comments in the latest push:

**1. Null safety + standard API (`HmacValidator.cs`)**
- Added a `string.IsNullOrEmpty(merchantSign)` guard so a null signature value returns `false` instead of throwing in `Encoding.UTF8.GetBytes`, restoring the previous null-safe behaviour.
- This library multi-targets `net462` and `netstandard2.0`, where `CryptographicOperations.FixedTimeEquals` isn't available. So I took Gemini's recommended alternative for that case: made the existing `HmacValidatorUtility.TimeSafeEquals` `internal` and reuse it here, instead of duplicating the comparison. Both webhook signature paths now go through the one helper.

**2. IL-inspection test (`HmacValidatorTest.cs`)**
- Agreed it's fragile across target frameworks / Debug-vs-Release / JIT-vs-AOT. Removed it and replaced it with a behavioural test (`TestIsValidHmac_NullSignatureValueReturnsFalse`) covering the null-value case above. The existing behavioural tests already cover valid/invalid signatures.

Happy to instead gate `CryptographicOperations.FixedTimeEquals` behind `#if NET8_0_OR_GREATER` for the net8 target if you'd prefer that over reusing the shared helper.

gcatanese:
🎊 Thanks for the contribution @tonghuaroot!

gcatanese (approved):

PR #1767, opened 2026-05-30
Fix ByteArrayConverter instantiated on every serialization call

### LINKED ISSUE #1739: Instance of new ByteArrayConverter().Read(...)  gets created every time.
nit:  This instance of `new ByteArrayConverter().Read(...) ` gets created every time. Perhaps we can do something like:

```csharp
// example
public class ThreeDSecureDataJsonConverter : JsonConverter<ThreeDSecureData> { 
       private readonly ByteArrayConverter byteArrayConverter;
       public ThreeDSecureDataJsonConverter()
       {
                byteArrayConverter = new ByteArrayConverter();
        }
}
```

_Originally posted by @Kwok-he-Chu in https://github.com/Adyen/adyen-dotnet-api-library/pull/1688#discussion_r3324189554_

gcatanese:
## Description
`ByteArrayConverter` was being instantiated inline (`new ByteArrayConverter().Read(...)` / `new ByteArrayConverter().Write(...)`) on every JSON serialization and deserialization of `byte[]` fields. Since the class is completely stateless, there is no reason to allocate a new instance per call.

This PR introduces a singleton `ByteArrayConverter.Instance` and updates the Mustache template and one example generated class (`Checkout/ThreeDSecureData`) to use it.

**Impact:** 9 generated model classes are affected on regeneration (those with `byte[]` fields). No other classes are touched by the template change.

## Tested scenarios
- Serialization roundtrip of `ThreeDSecureData` with all three `byte[]` fields (`Cavv`, `TokenAuthenticationVerificationValue`, `Xid`)
- Null `byte[]` field serialization
- Deserialization from raw JSON string

## Fixed issue
Fixes #1739

### DIFF
--- a/Adyen.Test/Core/Converters/ByteArrayTest.cs
+++ b/Adyen.Test/Core/Converters/ByteArrayTest.cs
@@ -8,7 +8,7 @@ namespace Adyen.Test.Core.Converters
     [TestClass]
     public class ByteArrayTest
     {
-        private readonly ByteArrayConverter _converter = ByteArrayConverter.Instance;
+        private readonly ByteArrayConverter _converter = new ByteArrayConverter();
         
         [TestMethod]
         public void Given_ByteArray_When_Write_Then_Result_Should_Write_UTF8_String()
--- a/Adyen.Test/Core/Converters/ByteArrayTest.cs
+++ b/Adyen.Test/Core/Converters/ByteArrayTest.cs
@@ -8,7 +8,7 @@ namespace Adyen.Test.Core.Converters
     [TestClass]
     public class ByteArrayTest
     {
-        private readonly ByteArrayConverter _converter = new ByteArrayConverter();
+        private readonly ByteArrayConverter _converter = ByteArrayConverter.Instance;
         
         [TestMethod]
         public void Given_ByteArray_When_Write_Then_Result_Should_Write_UTF8_String()
--- a/Adyen.Test/Core/Serialization/ByteArrayCoverageTests.cs
+++ b/Adyen.Test/Core/Serialization/ByteArrayCoverageTests.cs
@@ -22,7 +22,7 @@ namespace Adyen.Test.Core.Serialization
         private static JsonSerializerOptions BuildDiOptions()
         {
             var options = new JsonSerializerOptions();
-            options.Converters.Add(new ByteArrayConverter());
+            options.Converters.Add(ByteArrayConverter.Instance);
             options.Converters.Add(new ThreeDSecureDataJsonConverter());
             return options;
         }
--- a/Adyen.Test/Core/Serialization/ByteArrayCoverageTests.cs
+++ b/Adyen.Test/Core/Serialization/ByteArrayCoverageTests.cs
@@ -22,7 +22,7 @@ namespace Adyen.Test.Core.Serialization
         private static JsonSerializerOptions BuildDiOptions()
         {
             var options = new JsonSerializerOptions();
-            options.Converters.Add(ByteArrayConverter.Instance);
+            options.Converters.Add(new ByteArrayConverter());
             options.Converters.Add(new ThreeDSecureDataJsonConverter());
             return options;
         }
--- a/Adyen/Core/Converters/ByteArrayConverter.cs
+++ b/Adyen/Core/Converters/ByteArrayConverter.cs
@@ -10,9 +10,7 @@ namespace Adyen.Core.Converters
     public sealed class ByteArrayConverter : JsonConverter<byte[]>
     {
         public static ByteArrayConverter Instance { get; } = new();
-
-        private ByteArrayConverter() { }
-
+        
         /// <summary>
         /// Reads a byte array during deserialization.
         /// </summary>
--- a/Adyen/Core/Converters/ByteArrayConverter.cs
+++ b/Adyen/Core/Converters/ByteArrayConverter.cs
@@ -7,9 +7,11 @@ namespace Adyen.Core.Converters
     /// <summary>
     /// JsonConverter for byte arrays.
     /// </summary>
-    public class ByteArrayConverter : JsonConverter<byte[]>
+    public sealed class ByteArrayConverter : JsonConverter<byte[]>
     {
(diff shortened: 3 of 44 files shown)


Adyen/Core/Converters/ByteArrayConverter.cs
```diff
@@ -9,6 +9,8 @@ namespace Adyen.Core.Converters
     /// </summary>
     public class ByteArrayConverter : JsonConverter<byte[]>
     {
+        public static readonly ByteArrayConverter Instance = new();
```
Kwok-he-Chu: When I see these static method, **thread-safety** popped into my mind to ensure that this is completely stateless.


1. To address that this cannot be changed, **consider:** making the constructor private (= one instance) in this case and marking this class as `sealed` (because we don't want consumers to enherit this class).

```csharp
public sealed class ByteArrayConverter : JsonConverter<byte[]>
{ //...

 private ByteArrayConverter() { } // private constructor to force the singleton
/// ...
}
```

2. I'm used to writing these properties as follows - Did some diggin', taken from [source](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/property) from the official .NET microsoft design guidelines
a. "Properties should be seen as smart fields."
b. "DO create get-only properties if the caller should not be able to change the value of the property."

**Recommendation, change to:**
```csharp
  public static ByteArrayConverter Instance { get; } = new();
```

under the hood it generates:
```csharp
private ByteArrayConverter _instance;
public static ByteArrayConverter Instance
{
    get
    {
        return _instance;
    }
}
```

Kwok-he-Chu: _Note:_ for singletons specifically, I like the `Instance`-keyword. But I have also seen `Default` for a default implementation (configuration).

With the added `private ByteArrayConverter() { }`, I recommend that the SDK should keep the name `Instance`. Let's make sure this is used consistently across the entire SDK

Kwok-he-Chu: @gcatanese Up to you whether you want this change; it depends on what you want-



the line `private ByteArrayConverter() { } // private constructor to force the singleton` prevents consumer instantiation using: 
`new JsonOptions().Add(new ByteArrayConverter());` // wont be available

Adyen/Core/Converters/ByteArrayConverter.cs
```diff
@@ -7,8 +7,12 @@ namespace Adyen.Core.Converters
     /// <summary>
     /// JsonConverter for byte arrays.
     /// </summary>
-    public class ByteArrayConverter : JsonConverter<byte[]>
+    public sealed class ByteArrayConverter : JsonConverter<byte[]>
     {
+        public static ByteArrayConverter Instance { get; } = new();
+
+        private ByteArrayConverter() { }
```
Kwok-he-Chu: [Anchor (see comment)](https://github.com/Adyen/adyen-dotnet-api-library/pull/1767#discussion_r3333106245), reverting the change and allow public constructor instead.

This allows the consumer of the SDK to control the lifetime of the ByteArrayConverter by using `new ByteArrayConverter()`, this wouldn't be a breaking change then.

Kwok-he-Chu (changes requested):

Adyen/Core/Converters/ByteArrayConverter.cs
Kwok-he-Chu: Reason: Let's be pragmatic, so we don't change the entire SDK enforcing this approach - It doesn't align with how the current SDK **injects** the other Converters (see: https://github.com/Adyen/adyen-dotnet-api-library/blob/v34.0.3/Adyen/Checkout/Client/HostConfiguration.cs#L47-L69), e.g.

```csharp
                _jsonOptions.Converters.Add(new DateOnlyNullableJsonConverter()); // ... The "Instance" doesn't make sense here
```

gcatanese: @Kwok-he-Chu ✅

Kwok-he-Chu (dismissed):

Adyen.Test/Core/Serialization/ByteArrayCoverageTests.cs
```diff
@@ -22,7 +22,7 @@ public class ByteArrayCoverageTests
         private static JsonSerializerOptions BuildDiOptions()
         {
             var options = new JsonSerializerOptions();
-            options.Converters.Add(new ByteArrayConverter());
+            options.Converters.Add(ByteArrayConverter.Instance);
```
Kwok-he-Chu: should be: `options.Converters.Add(new ByteArrayConverter());`

Adyen.Test/Core/Converters/ByteArrayTest.cs
```diff
@@ -8,7 +8,7 @@ namespace Adyen.Test.Core.Converters
     [TestClass]
     public class ByteArrayTest
     {
-        private readonly ByteArrayConverter _converter = new ByteArrayConverter();
+        private readonly ByteArrayConverter _converter = ByteArrayConverter.Instance;
```
Kwok-he-Chu: Use:  `private readonly ByteArrayConverter _converter = new ByteArrayConverter();` here since we're instantiating it manually

Adyen.Test/Core/Serialization/ByteArrayCoverageTests.cs
Kwok-he-Chu: See: https://github.com/Adyen/adyen-dotnet-api-library/pull/1767/changes#r3341380022

Adyen.Test/Core/Converters/ByteArrayTest.cs
gcatanese: Done

Adyen.Test/Core/Serialization/ByteArrayCoverageTests.cs
gcatanese: Done

Kwok-he-Chu (approved):

PR #1768, opened 2026-05-30
Revise byte[] deserialization to use standard STJ delegation

### LINKED ISSUE #1738: Revise handling of byte[] deserialization
I see two main ways to handle this `byte[]` deserialization. Ideally, the SDK should keep using the following line, allowing consumers to register their own byte[] converters passed to the `JsonSerializerOptions`.

____

**Option 1:** Consumer-facing registration
```csharp
// consumer:
services.AddControllers()
    .AddJsonOptions(options =>
    {
        options.JsonSerializerOptions.Converters.Add(new ByteArrayConverter()); // requires consumer to specify the converter
    });

// SDK:
case "cavv":
    cavv = new Option<byte[]?>(
        JsonSerializer.Deserialize<byte[]>(
            ref utf8JsonReader,
            jsonSerializerOptions)); // preferred way and use: byte[]
    break;
```

**Option 2:** Do an explicit look-up, and check if if we're using System.Text.Json or our own. We can enforce it here // note: pseudo code, needs to be tested*
_Example:_
```csharp
/// SDK
bool isCustomConverter =
    options.Converters.Any(c => c.CanConvert(typeof(byte[])));             // if it's false we use "System.Text.Json.Serialization.Converters.ByteArrayConverter"  , not sure if I have to check the fully-qualified name here

// Then assign or instantiate the right converter
i
              JsonConverter<byte[]> byteArrayConverter;

if (!isCustomConverter)
    byteArrayConverter = (JsonConverter<byte[]>)jsonSerializerOptions.GetConverter(typeof(byte[]));
else
    byteArrayConverter = new Adyen.Core.Converters.ByteArrayConverter();

case "cavv":
    cavv = new Option<byte[]?>(
        byteArrayConverter.Read(
            ref utf8JsonReader,
            typeof(byte[]),
            jsonSerializerOptions));
    break;
```

_Originally posted by @Kwok-he-Chu in https://github.com/Adyen/adyen-dotnet-api-library/pull/1688#discussion_r3324181853_

gcatanese:
## Description

Use `JsonSerializer.Deserialize<byte[]>` / `JsonSerializer.Serialize` for all `byte[]` fields in generated model converters, instead of hardcoding `ByteArrayConverter` directly. This delegates byte[] handling to whatever converter is registered in `JsonSerializerOptions`, restoring standard STJ extensibility.

**Key changes:**
- `ByteArrayConverter` is no longer called directly in generated converters — `JsonSerializerOptions` is respected
- `ByteArrayConverter` remains registered in `HostConfiguration` for backward compatibility in DI mode
- In bare mode (no options), STJ's default base64 behavior is used — aligning with the Java SDK's Jackson default handling
- 10 generated model files updated; `JsonConverter.mustache` template updated for future regeneration

## Tested scenarios
- DI mode: `ByteArrayConverter` from options is used (existing behavior preserved)
- Bare mode: STJ default base64 decode/encode (aligns with Java SDK)
- Consumer override: custom `byte[]` converter registered in options is now honored
- Roundtrip serialization for all affected models

## Fixed issue
Fixes #1738

### DIFF
--- a/Adyen.Test/Checkout/CheckoutPaymentsBareSerializationTests.cs
+++ b/Adyen.Test/Checkout/CheckoutPaymentsBareSerializationTests.cs
@@ -110,8 +110,10 @@ namespace Adyen.Test.Checkout
         [TestMethod]
         public void Given_ThreeDSecureData_When_BareSerialize_Then_ByteArray_IsCorrect()
         {
-            byte[] base64Bytes = Encoding.ASCII.GetBytes(Convert.ToBase64String(Encoding.ASCII.GetBytes("Bytes-To-Be-Encoded")));
-            var threeDSecure = new ThreeDSecureData { Cavv = base64Bytes, Xid = base64Bytes };
+            // STJ default: byte[] is base64-encoded. Encoding.UTF8.GetBytes("Bytes-To-Be-Encoded")
+            // base64-encodes to "Qnl0ZXMtVG8tQmUtRW5jb2RlZA==".
+            byte[] rawBytes = Encoding.UTF8.GetBytes("Bytes-To-Be-Encoded");
+            var threeDSecure = new ThreeDSecureData { Cavv = rawBytes, Xid = rawBytes };
 
             string result = JsonSerializer.Serialize(threeDSecure);
 
@@ -121,12 +123,13 @@ namespace Adyen.Test.Checkout
         [TestMethod]
         public void Given_ThreeDSecureData_When_BareDeserialize_Then_ByteArray_IsCorrect()
         {
+            // STJ default: base64-decodes the JSON string. "Qnl0ZXMtVG8tQmUtRW5jb2RlZA==" decodes to "Bytes-To-Be-Encoded".
             string json = "{\"cavv\":\"Qnl0ZXMtVG8tQmUtRW5jb2RlZA==\",\"xid\":\"Qnl0ZXMtVG8tQmUtRW5jb2RlZA==\"}";
 
             var result = JsonSerializer.Deserialize<ThreeDSecureData>(json);
 
-            Assert.AreEqual("Qnl0ZXMtVG8tQmUtRW5jb2RlZA==", Encoding.ASCII.GetString(result.Cavv));
-            Assert.AreEqual("Qnl0ZXMtVG8tQmUtRW5jb2RlZA==", Encoding.ASCII.GetString(result.Xid));
+            Assert.AreEqual("Bytes-To-Be-Encoded", Encoding.UTF8.GetString(result.Cavv));
+            Assert.AreEqual("Bytes-To-Be-Encoded", Encoding.UTF8.GetString(result.Xid));
         }
 
         #endregion
--- a/Adyen.Test/Checkout/CheckoutPaymentsBareSerializationTests.cs
+++ b/Adyen.Test/Checkout/CheckoutPaymentsBareSerializationTests.cs
@@ -323,6 +323,107 @@ namespace Adyen.Test.Checkout
         }
 
         [TestMethod]
+        public void Given_PaymentRequest_With_MpiData_When_BareSerialize_Then_CavvIsBase64Encoded()
+        {
+            // "3q2+78r+ur7erb7vyv66vv////8=" is the CAVV from the Adyen 3DS mpiData docs sample.
+            byte[] cavvBytes = Convert.FromBase64String("3q2+78r+ur7erb7vyv66vv////8=");
+            var request = new PaymentRequest
+            {
+                MerchantAccount = "YOUR_MERCHANT_ACCOUNT",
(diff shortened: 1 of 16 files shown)


Adyen.Test/Checkout/CheckoutPaymentsBareSerializationTests.cs
```diff
@@ -110,8 +110,10 @@ public void Given_DeviceRenderOptions_When_BareSerialize_Then_SdkUiType_Contains
         [TestMethod]
         public void Given_ThreeDSecureData_When_BareSerialize_Then_ByteArray_IsCorrect()
         {
-            byte[] base64Bytes = Encoding.ASCII.GetBytes(Convert.ToBase64String(Encoding.ASCII.GetBytes("Bytes-To-Be-Encoded")));
```
Kwok-he-Chu, line 113: **[Breaking change??]** This test changes what we're assigning to the `ThreeDSSecureData` model, e.g. from base64Bytes to rawBytes (see: `var threeDSecure = new ThreeDSecureData { Cavv = rawBytes, Xid = rawBytes };`)

**Address:** consumers that directly use the `ThreeDSSecureData` when they upgrade.

Can you confirm what the Adyen API outputs and whether this is a breaking change?

_Example below-_

**rawBytes:**
1. `byte[] rawBytes = Encoding.UTF8.GetBytes("Bytes-To-Be-Encoded");`

```
 // rawBytes:
  {byte[19]}
    [0]: 66
    [1]: 121
    [2]: 116
    [3]: 101
    [4]: 115
    [5]: 45
    [6]: 84
    [7]: 111
    [8]: 45
    [9]: 66
    [10]: 101
    [11]: 45
    [12]: 69
    [13]: 110
    [14]: 99
    [15]: 111
    [16]: 100
    [17]: 101
    [18]: 100
```

---

**base64bytes:**

2. `byte[] base64Bytes = Encoding.ASCII.GetBytes(Convert.ToBase64String(Encoding.ASCII.GetBytes("Bytes-To-Be-Encoded")));`

```
  //base64Bytes
  {byte[28]}
    [0]: 81
    [1]: 110
    [2]: 108
    [3]: 48
    [4]: 90
    [5]: 88
    [6]: 77
    [7]: 116
    [8]: 86
    [9]: 71
    [10]: 56
    [11]: 116
    [12]: 81
    [13]: 109
    [14]: 85
    [15]: 116
    [16]: 82
    [17]: 87
    [18]: 53
    [19]: 106
    [20]: 98
    [21]: 50
    [22]: 82
    [23]: 108
    [24]: 90
    [25]: 65
    [26]: 61
    [27]: 61
```

Kwok-he-Chu (changes requested):

Adyen.Test/Checkout/CheckoutPaymentsBareSerializationTests.cs
```diff
@@ -110,8 +110,10 @@ public void Given_DeviceRenderOptions_When_BareSerialize_Then_SdkUiType_Contains
...
-            var threeDSecure = new ThreeDSecureData { Cavv = base64Bytes, Xid = base64Bytes };
+            // STJ default: byte[] is base64-encoded. Encoding.UTF8.GetBytes("Bytes-To-Be-Encoded")
+            // base64-encodes to "Qnl0ZXMtVG8tQmUtRW5jb2RlZA==".
+            byte[] rawBytes = Encoding.UTF8.GetBytes("Bytes-To-Be-Encoded");
```
Kwok-he-Chu, line 115: see https://github.com/Adyen/adyen-dotnet-api-library/pull/1768/changes#r3333329570

Adyen.Test/Checkout/CheckoutPaymentsBareSerializationTests.cs
gcatanese: @Kwok-he-Chu 

I have looked/compared with Java library `byte[]` handling as they should behave the same (the assumption is that the Java library is performing this correctly).

The Java `ThreeDSecureData` declares `cavv`, `xid`, and `tokenAuthenticationVerificationValue` as `byte[]` fields with Jackson `@JsonProperty` annotations. Jackson's default behavior for `byte[]` is standard base64 encode/decode -- exactly what your .NET PR #1768 aligns bare mode to.

The Java library does not have explicit tests that validate the byte[] base64 roundtrip through Jackson on ThreeDSecureData or similar models. 

Bottom line: The Java library implicitly relies on Jackson's default base64 handling for `byte[]` without testing it explicitly. The .NET PR is correct in aligning with this behavior, but there are no Java tests to copy. The tests you already have in the .NET PR (bare serialization/deserialization of `ThreeDSecureData` with base64 assertions) are actually more thorough than what exists in Java.

I would then leave the PR as-is. Meanwhile I will try to collect the actual response payload from API responses on TEST.

PR #1859, opened 2026-09-14
Document live endpoint configuration

gcatanese:
**Description**
Add a Checkout API Live configuration example showing how to set `LiveEndpointUrlPrefix` and the resulting endpoint format.

**Tested scenarios**
- Verified Markdown changes with `git diff --check`.

**Fixed issue**: N/A


andrej-gajduk-adyen (approved):
LGTM

